PR #552, opened 2026-06-13
fix(ci): use @v1 managed tag for all projectbluefin/actions refs

castrojo:
## Problem

All `projectbluefin/actions` workflow refs in this repo were SHA-pinned (e.g. `@c3669632...` with a `# v1` comment). This causes a reliability problem:

- Every fix merged into `projectbluefin/actions` requires a separate Renovate bump PR in every consumer repo before it takes effect
- That lag is typically **days**
- Tonight this bit us directly: a 6-day-old `--auto` bug in the shared automerge workflow blocked a release here because the SHA hadn't been bumped yet

## Fix

Replace all 16 SHA pins with `@v1`, the managed tag that `projectbluefin/actions` already maintains for stable consumers. Changes in shared actions now propagate instantly to all repos using `@v1`.

**AGENTS.md already documents this policy:**
> `projectbluefin/` refs (`@v1`, `@main`) are intentional managed tags and are exempted.

## Pre-commit enforcement

Added a `no-sha-pins-for-internal-actions` hook to `.pre-commit-config.yaml` alongside the existing `no-floating-action-tags` hook. It blocks any future `uses: …projectbluefin/…@<40-char-SHA>` from being committed.

## Files changed

- 15 workflow files — SHA pins → `@v1`
- `.pre-commit-config.yaml` — new enforcement hook

Only `projectbluefin/actions` refs were touched. All external action pins (e.g. `actions/checkout@SHA`) are unchanged.

### DIFF
--- a/.github/workflows/build-image-testing.yml
+++ b/.github/workflows/build-image-testing.yml
@@ -41,7 +41,7 @@ jobs:
       should_build: ${{ steps.detect.outputs.should_build }}
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
-      - uses: projectbluefin/actions/bootc-build/detect-changes@6231015b336556d2ff0adc1d1e59514bf19dcb42 # v1
+      - uses: projectbluefin/actions/bootc-build/detect-changes@v1
         id: detect
 
   build-image-testing:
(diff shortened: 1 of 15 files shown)


hanthor (changes requested):
## Adversarial review — fix needed

### [BLOCKER] Targets `main`, must target `testing`
The `Check PR base branch` CI check fails. All bluefin PRs must target `testing`. Re-open with `--base testing`.

### [HIGH] Promotion collision risk
Promotion PR #545 contains SHA-pinned refs in the same workflow files this PR converts to `@v1`. If #545 merges first, it reintroduces SHA pins that this PR then needs to fix again. **Recommendation:** merge this PR **before** the next promotion cycle.

### [MEDIUM] Lost semantic comment
`consumer-validate-generate-release-notes.yml` — the `# consumer-validation` comment on the old SHA pin distinguished the purpose of this override. The `@v1` conversion loses that context. Consider adding a comment explaining why this workflow uses `@v1` specifically.

PR #574, opened 2026-06-15
fix(ci): publish :testing stream tag on push to testing branch

castrojo:
## Problem

`publish_stream_tag: "false"` has been in place since the repo was bootstrapped. This means every push to `testing` builds the image but never writes the `:testing` OCI tag — so `ghcr.io/projectbluefin/bluefin:testing` does not exist / never updates.

## Fix

Flip to `"true"`. Every push to `testing` (whether from a mergeraptor automerge or a manual commit) will now publish:
- `ghcr.io/projectbluefin/bluefin:testing`
- `ghcr.io/projectbluefin/bluefin:testing-nvidia`

## Related

Part of making the `testing` → `main` promotion loop fully automatic. The companion branch-protection fix (adding `Unit tests` as a required check on `testing`) is being applied separately via the GitHub API.

### DIFF
--- a/.github/workflows/build-image-testing.yml
+++ b/.github/workflows/build-image-testing.yml
@@ -70,5 +70,5 @@ jobs:
           || '["main", "nvidia"]'
         }}
       stream_name: testing
-      publish_stream_tag: "false"
+      publish_stream_tag: "true"
       pr_number: ${{ inputs.pr_number }}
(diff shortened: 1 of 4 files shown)


castrojo:
Closing — this was wrong. The `:testing` tag is correctly managed by `post-testing-e2e.yml`, which only promotes a digest to `:testing` *after* e2e smoke+common passes. Changing `publish_stream_tag` to `true` here would bypass that e2e gate and let unverified images land as `:testing`.

PR #712, opened 2026-06-24
fix(release): pass fast_forward_sha to reusable promotion

castrojo:
## Summary

projectbluefin/actions `reusable-execute-release.yml` adds an optional `fast_forward_sha` input that callers should pass to avoid the unsafe `github.sha` fallback and a deprecation warning. Future v2 will make it required.

This PR plumbs `${{ github.event.workflow_run.head_sha || github.sha }}` — `workflow_run.head_sha` is the verified source SHA; `github.sha` is the correct fallback for the `push: main` trigger path bluefin uses.

## Coordination

Depends on projectbluefin/actions PR (backward-compat, safe to merge here first if needed).

Assisted-by: Claude Opus 4.7 via GitHub Copilot


castrojo:
Recreating with correct base branch (testing, not main). See replacement PR.

PR #892, opened 2026-07-23
feat: Apple Silicon (Asahi) flavor — bluefin-asahi (aarch64)

hanthor:
Makes Bluefin bootable on M1/M2 Macs via the Fedora Asahi Remix stack (@asahi COPRs) — the companion to projectbluefin/bluefin-lts#448, both ported from the tuna-os overlay that produced the first promoted TunaOS asahi image earlier today (35/35 on the Asahi verify harness; installer payload boots U-Boot EFI → GNOME userspace in QEMU).

**Kernel path:** the asahi flavor branches out of `04-install-kernel-akmods.sh` before any akmods machinery: `@asahi/fedora-remix-branding` COPR → `asahi-repos` → `kernel-16k` + `asahi-platform-metapackage` + `alsa-ucm-asahi`/`tiny-dfr` + `uboot-images-armv8` (apple_m1 U-Boot payload) + `asahi-fwupdate`/`dracut-asahi`/`update-m1n1` — the same set Fedora Asahi Remix atomic desktops use. Hard asserts on the 16K asahi kernel and Apple DTBs in the modules tree. The existing `19-initramfs.sh` dracut rebuild picks up dracut-asahi's ESP vendor-firmware modules.

**boot.bin lifecycle:** ships `asahi-bootbin-sync` (pinned from tuna-os/bootc-installer-asahi) — bootc deploys never run package scriptlets, so update-m1n1 otherwise never re-runs after `bootc upgrade`. Apple-DT-gated; no-op off Apple hardware.

**Wiring:** `asahi` in the Justfile flavor map; a **separate reusable-build call** in build-image-testing.yml with `architecture: ["aarch64"]` so the x86_64 matrix never sees the flavor (the script hard-guards non-aarch64 anyway). `ghcr.io/ublue-os/akmods` is already multiarch, so shared plumbing is unaffected; nvidia/zfs akmods simply don't apply on Apple Silicon.

**Open questions for review:**
- Flavor naming: this follows the repo's flavor→image-name convention (`bluefin-asahi`, like `bluefin-nvidia`). The TunaOS-side rule is 'asahi is a tag suffix' — happy to rework if you want tag-suffix semantics here instead.
- First aarch64 leg for this repo: the asahi call is scoped so nothing else changes, but the reusable workflow's aarch64 runner path gets its first bluefin exercise.
- No ISO/installer path — Apple Silicon cannot external-boot; installs go through the asahi-installer flow (tuna-os/bootc-installer-asahi).

Part of the Apple Silicon effort tracked at tuna-os/tunaOS#781.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018bGymcRkVtV7DPToWDjZyZ

### DIFF
--- a/.github/workflows/build-image-testing.yml
+++ b/.github/workflows/build-image-testing.yml
@@ -72,3 +72,28 @@ jobs:
       stream_name: testing
       publish_stream_tag: "false"
       pr_number: ${{ inputs.pr_number }}
+
+  build-image-testing-asahi:
+    name: Build Testing Images (Asahi)
+    needs: [detect-changes]
+    # Apple Silicon flavor: aarch64 only (the build script hard-guards),
+    # separate call so the main x86_64 matrix never sees the asahi flavor.
+    if: |
+      always() &&
+      (github.event_name != 'workflow_dispatch' || github.ref_name == 'main' || inputs.pr_number != '') &&
+      (github.event_name != 'pull_request' ||
+       needs.detect-changes.outputs.should_build == 'true')
+    uses: projectbluefin/actions/.github/workflows/reusable-build.yml@v1
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+      attestations: write
+      security-events: write
+      issues: write
+    with:
+      image_flavors: '["asahi"]'
+      architecture: '["aarch64"]'
+      stream_name: testing
+      publish_stream_tag: "false"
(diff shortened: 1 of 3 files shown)


castrojo:
Strong -1 on adding more COPRs. Let's just make the Dakota image the top tier one since we have bst there and don't need to add third party stuff.

PR #911, opened 2026-07-28
feat(ci): add BATS coverage reporting artifacts

castrojo:
## Summary
- install `kcov` in PR validation unit-test job
- run BATS tests with coverage instrumentation for `build_files/` and `system_files/`
- upload coverage output as a `bats-kcov-report` artifact
- keep existing TAP artifact upload and document this behavior in the CI skill

## Why
Issue #518 reports that Bluefin has BATS unit tests but no shell coverage reporting. This adds CI-visible coverage output without requiring external secrets or token setup.

Fixes projectbluefin/bluefin#518

## Validation
- `python3 -m compileall tests`
- `just check` *(blocked locally: `just` not installed in this environment)*
- `pre-commit run --all-files` *(blocked locally because `just`/toolchain prerequisites are unavailable)*

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -153,14 +153,20 @@ jobs:
           export PATH="${GITHUB_WORKSPACE}/tests/coverage/bin:${PATH}"
           mkdir -p "${KCOV_RAW_DIR}"
 
+          # The instrumented rerun exists to collect coverage; the "Run unit
+          # tests" step above owns pass/fail, so a failing test must not mask
+          # the coverage report.
+          coverage_status=0
           kcov \
             --bash-method=DEBUG \
             --bash-parse-files-in-dir="${INCLUDE_PATHS}" \
             --include-path="${INCLUDE_PATHS}" \
             --exclude-path="${EXCLUDE_PATHS}" \
             "${KCOV_RAW_DIR}/top" \
-            bash -c 'exec bats --formatter tap tests/unit/ > coverage/bats-coverage-run.tap 2>&1'
-          tail -n 20 coverage/bats-coverage-run.tap
+            bash -c 'exec bats --formatter tap tests/unit/ > coverage/bats-coverage-run.tap 2>&1' \
+            || coverage_status=$?
+          echo "instrumented bats exit status: ${coverage_status}"
+          grep -c '^ok' coverage/bats-coverage-run.tap
           python3 tests/coverage/merge_kcov.py \
             --raw-dir "${KCOV_RAW_DIR}" \
             --output-dir coverage/kcov \
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -79,18 +79,31 @@ jobs:
     needs: [check-base-branch]
     if: always() && (needs.check-base-branch.result == 'success' || needs.check-base-branch.result == 'skipped')
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 15
     permissions:
       contents: read
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


hanthor (changes requested):
The required unit-test workflow is red because `sudo apt-get install -y bats kcov` fails on the Ubuntu 24.04 runner (`E: Unable to locate package kcov`). Please install kcov from an available, pinned source (or use a runner/image that provides it) before enabling the coverage step.

bketelsen:
I opened castrojo/bluefin-1#1 against this PR's head branch. It fixes the unavailable Ubuntu 24.04 `kcov` package by building v43 from a checksum-verified pinned commit.

Validation exposed a second issue: wrapping only the top-level `bats` process produces a 0% report because the scripts under test run in child shells. The stacked PR source-maps those child runs back to the original repository files and emits merged HTML/Cobertura output.

castrojo:
@hanthor thanks — your blocking review was correct, and validation turned up a second (worse) defect. Both are fixed, and this branch now also carries @bketelsen's work from castrojo/bluefin-1#1.

**1. `kcov` is not installable on Ubuntu 24.04 (your review).** Confirmed: `apt-get install -y kcov` in an `ubuntu:24.04` container gives `E: Unable to locate package kcov` — kcov was dropped from noble. Fixed by building kcov **v43 from a pinned source commit** (`a39874f938ce13f7a65f253120d1ec946b349ffe`) whose release archive SHA-256 is verified with `sha256sum --check` before it is unpacked. No PPA, no floating URL, no unverified binary. The build output is cached (`actions/cache`, key `kcov-v43-a39874f-...`) so only the first run pays the compile cost; job timeout raised to 25 minutes to cover a cold build.

**2. The report was 0%.** Wrapping only the top-level `bats` process produces a valid-looking but empty report, because every test executes the script under test in a child shell (`run bash "${PATCHED_SCRIPT}"`) against a patched sandbox copy. Fixed by routing those child `bash <script>` calls through `tests/coverage/bin/bash`, which starts a nested kcov run and records the sandbox copy's original `build_files/`/`system_files/` path; `tests/coverage/merge_kcov.py` then merges those child hits into kcov's pre-parsed source inventory and emits HTML + Cobertura + a step-summary table. It hard-fails if zero lines were executed, so a silent 0% report can never ship again.

**3. (found during validation) the coverage step deadlocked.** With the instrumented run streaming the full TAP log through kcov's captured stdout pipe, kcov stops draining once the pipe buffer fills and the job hangs forever — I reproduced this twice locally (hung at ~96 traced child runs, no progress for 15+ min). Fixed by redirecting the coverage run's BATS output to a file. That also removed the backpressure: the instrumented suite now finishes in **~17 seconds** instead of stalling.

**Evidence (Ubuntu 24.04 container, pinned kcov v43, exact CI command):**

```
BATS shell coverage: 269/564 lines (47.7%)

| File | Covered | Valid | Coverage |
| build_files/base/00-image-info.sh          |  39 |  40 |  97.5% |
| build_files/base/03-packages.sh            |  26 |  26 | 100.0% |
| build_files/base/05-override-install.sh    |  21 |  21 | 100.0% |
| build_files/base/17-cleanup.sh             |  28 |  29 |  96.6% |
| build_files/shared/build-gnome-extensions.sh | 22 | 22 | 100.0% |
| system_files/.../11-framework-ucsi-workaround.sh | 17 | 23 | 73.9% |
| ... 27 files total
```

Verified identical BATS pass/fail results with and without kcov, so instrumentation does not change test outcomes. Known limitation, documented in `docs/skills/ci/SKILL.md`: tests that `source` a library directly into the BATS process (e.g. `package-lib`, `validate-repos`) are not traced.

Also closed #974, which was a byte-identical resubmit of this PR targeting `main` in violation of the testing-first flow.

PR #912, opened 2026-07-28
fix(sleep): disable psmouse wakeup on HP ZBook Ultra G1a + add missing 20-framework.sh hook

### LINKED ISSUE #383: AMD s2idle never reaches deepest state on HP ZBook Ultra G1a — atkbd wakeup blocks PC10/S0i3
A user reported poor battery life on the HP ZBook Ultra G1a 14" running Bluefin (live image, kernel 6.19.14-101.fc44.x86_64). Log analysis identified the root cause as a missing kernel quirk. The CachyOS kernel works correctly for them, which points to a kernel-level fix being the right solution.

## Hardware

- **Machine:** HP HP ZBook Ultra G1a 14 inch Mobile Workstation PC
- **Board name (DMI):** `8D01`
- **SoC:** AMD (amdgpu, AMD-Vi, `amd_pmc AMDI000B:00`)
- **BIOS:** HP X89 Ver. 01.03.00 (2025-04-25) — first release firmware
- **Kernel:** 6.19.14-101.fc44.x86_64
- **Sleep mode:** s2idle (`Low-power S0 idle used by default for system suspend`)

## Symptom

Every single suspend cycle logs on resume:

```
amd_pmc AMDI000B:00: Last suspend didn't reach deepest state
```

## Measured Impact

Sleep drain measured over a 70-minute suspend (09:05:48 → 10:16:31):

| | |
|---|---|
| Energy at sleep | 64,516,000 µWh |
| Energy at wake | 62,731,000 µWh |
| Consumed | 1,785,000 µWh (1.785 Wh) |
| Average sleep power | **~1.51W** |

Expected for proper AMD s2idle with PC10: **0.1–0.5W**. This machine is consuming 3–15× more than it should during sleep.

## Root Cause

The PS/2 keyboard controller (`i8042`/`atkbd`) is registered as a wakeup source. The AMD PMC driver requires all devices to enter D3 before the SoC can reach PC10/S0i3. The atkbd wakeup registration prevents that transition, keeping the system in a shallow C-state throughout suspend.

This is the same class of bug already fixed upstream for a long list of Lenovo ThinkPads and other AMD laptops in `drivers/platform/x86/amd/pmc/pmc-quirks.c` via the `quirk_s2idle_spurious_8042` entry. The HP ZBook Ultra G1a (`8D01`) is simply not yet in that list.

## Why CachyOS Works

CachyOS ships kernel patches including AMD power management fixes that likely include the equivalent of `quirk_s2idle_spurious_8042` for this hardware, or carry the fix ahead of the Fedora kernel.

## Upstream Fix Needed

Add the HP ZBook Ultra G1a to the `fwbug_list` in `drivers/platform/x86/amd/pmc/pmc-quirks.c`:

```c
{
    .ident = "HP ZBook Ultra G1a 14",
    .driver_data = &quirk_s2idle_spurious_8042,
    .matches = {
        DMI_MATCH(DMI_BOARD_VENDOR, "HP"),
        DMI_MATCH(DMI_BOARD_NAME, "8D01"),
    }
},
```

This is a one-liner DMI entry, the same pattern used for every other affected laptop in that file. The patch should go to `linux-kernel@vger.kernel.org` with `platform-driver-x86` maintainers cc'd.

**Patch:** https://gist.github.com/castrojo/647c8e1c99e54bc146021901b1aebaa8

## Short-term Workaround (already carried in common)

`projectbluefin/common` now ships `usr/lib/udev/rules.d/60-amd-s2idle-fixes.rules` which disables atkbd wakeup for board `8D01` at boot. This is the userspace equivalent of `quirk_s2idle_spurious_8042` and is intended as a bridge until the upstream kernel patch lands. Once the fix is in a shipped Fedora kernel, the board entry should be removed from that file.

## Prior Art

- Upstream `pmc-quirks.c` already lists ~10 Lenovo boards with the identical fix
- Framework 13 Ryzen 7040: same fix applied in `10-framework.sh` setup hook (removed after BIOS 3.09)
- Bazzite applies the same atkbd udev rule for Framework AMD, also as a per-vendor script
- No distro has a systematic userspace solution — the kernel is the right place for this

castrojo:
## AMD s2idle / HP ZBook Ultra G1a — psmouse wakeup supplement

Closes #383

`common` ships `60-amd-s2idle-fixes.rules` which disables the `atkbd` (serio0) wakeup source on board `8D01`. However a commenter on #383 reported that even with `serio0` wakeup disabled, sleep drain remains at 2–3 %/h. The `amd_pmc` driver requires **all** PS/2 controller children to be in D3 before the SoC can enter PC10/S0i3, so `serio1` (psmouse) must also be disabled.

### Change 1 — `61-amd-s2idle-hp.rules`

Adds a udev rule to disable the psmouse (serio1) wakeup source on board `8D01`. Together with common's existing `60-amd-s2idle-fixes.rules` this covers the full i8042 child-device wakeup surface — the userspace equivalent of the single `quirk_s2idle_spurious_8042` kernel entry being submitted upstream.

File carries a removal note: drop once the kernel fix for `pmc-quirks.c` lands in a shipped Fedora kernel.

### Change 2 — `20-framework.sh` user-setup hook

`tests/unit/20-framework_test.bats` references `system_files/.../user-setup.hooks.d/20-framework.sh` but the file was never created, causing `bats tests/unit/` to fail with `No such file or directory`.

This adds the implementation:
- Detects Framework chassis via DMI `chassis_vendor`
- Guards for missing brew / non-writable Homebrew prefix
- Installs `fw-ectool` (Framework EC tool) if not already present
- Idempotent: prints `already installed, skipping` on repeat runs

All five bats scenarios pass against the new script.

## Testing

```bash
bash -n system_files/shared/usr/share/ublue-os/user-setup.hooks.d/20-framework.sh
# All 20-framework_test.bats scenarios pass manually (verified against test mocks)
```

The udev rule file is syntax-correct; no shell to lint.

### DIFF
--- a/system_files/shared/usr/lib/udev/rules.d/61-amd-s2idle-hp.rules
+++ b/system_files/shared/usr/lib/udev/rules.d/61-amd-s2idle-hp.rules
@@ -1,0 +1,20 @@
+# AMD s2idle deep-sleep supplement — HP ZBook Ultra G1a 14 (board 8D01)
+#
+# common already disables the PS/2 keyboard (serio/atkbd) via
+# 60-amd-s2idle-fixes.rules.  Field reports show that is not always
+# sufficient: the PS/2 mouse controller (serio/psmouse) can also block
+# the AMD SoC from entering PC10/S0i3, producing ~1.5 W sleep drain
+# instead of the expected 0.1–0.5 W.
+#
+# This rule disables psmouse wakeup to complete the fix begun in common.
+# Both rules together are the userspace equivalent of the single-device
+# quirk_s2idle_spurious_8042 kernel entry for this board.
+#
+# Remove this file once the upstream kernel fix for board 8D01 lands in a
+# shipped Fedora kernel (tracking: projectbluefin/bluefin#383).
+
+# HP ZBook Ultra G1a 14 — disable PS/2 mouse (serio1 / psmouse) wakeup
+ACTION=="add", SUBSYSTEM=="serio", DRIVERS=="psmouse", \
+  PROGRAM=="/bin/cat /sys/class/dmi/id/board_name", \
+  RESULT=="8D01", \
+  ATTR{power/wakeup}="disabled"
--- a/system_files/shared/usr/share/ublue-os/user-setup.hooks.d/20-framework.sh
+++ b/system_files/shared/usr/share/ublue-os/user-setup.hooks.d/20-framework.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/bash
+# Per-vendor user setup for Framework laptops.
+# Installs the Framework EC tool via Homebrew for hardware management.
+
+# shellcheck source=/dev/null
+source /usr/lib/ublue/setup-services/libsetup.sh
+
+version-script 20-framework user 1 || exit 0
+
+set -euo pipefail
+
+CHASSIS_VENDOR_PATH="/sys/devices/virtual/dmi/id/chassis_vendor"
+BREW_PREFIX="/home/linuxbrew/.linuxbrew"
+
+# Only run on Framework hardware
+[[ -r "${CHASSIS_VENDOR_PATH}" ]] || exit 0
+chassis_vendor="$(cat "${CHASSIS_VENDOR_PATH}")"
+[[ "${chassis_vendor}" == "Framework" ]] || exit 0
+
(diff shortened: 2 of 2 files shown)


hanthor (approved):
The hardware scope and idempotent setup hook look correct, and validation/unit tests are green. Approved.

PR #916, opened 2026-07-28
fix(lab-check): use GITHUB_TOKEN instead of MergeRaptor App token

castrojo:
Aligns with the bluefin-lts pattern: add `checks: write` to the workflow `permissions:` block and use `${{ github.token }}` directly. Removes the dependency on `actions/create-github-app-token` entirely, which was failing with 422 on every run (MergeRaptor installation hasn't approved `checks:write` via v3's fine-grained token scoping).

Closes #915

### DIFF
--- a/.github/workflows/lab-check.yml
+++ b/.github/workflows/lab-check.yml
@@ -6,6 +6,7 @@ on:
 
 permissions:
   contents: read
+  checks: write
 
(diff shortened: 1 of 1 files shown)


castrojo:
Superseded — updating PR #915 instead

PR #962, opened 2026-08-02
fix(image): include bootupd for bootc installs

castrojo:
## Summary

Add `bootupd` explicitly to the base Fedora package manifest so `bootc install` can install the bootloader during release-gate VM validation.

## Validation

- `python3 build_files/shared/read-packages build_files/packages/base.toml fedora | grep -Fx bootupd`
- `just check`
- `pre-commit run --all-files`


castrojo:
merging since we need this for release.

PR #913, opened 2026-07-28
ci: restore run-upgrade-test as promotion gate

castrojo:
## What does this change?

Restores `run-upgrade-test` as a promotion gate in `post-testing-e2e.yml`.

## Why?

Closes #400

**Root cause re-investigation:**

The issue attributed the lifecycle failure to GNOME 50 AT-SPI, but investigation of the actual run logs and testsuite source reveals:

- The lifecycle test suite is **pure SSH** — no GNOME session, no AT-SPI. The testsuite `e2e.yml` explicitly skips AT-SPI setup for lifecycle shards (`suite_dir != 'lifecycle'` conditionals). AT-SPI changes cannot affect lifecycle tests.
- The original failure (run [27075323936](https://github.com/projectbluefin/bluefin/actions/runs/27075323936)) showed a **VM boot/bootupd issue**, not an AT-SPI regression. The bootupd error was handled, but SSH timed out.
- The testsuite SHA has since been bumped to `f71d58d` (includes `ghcr.io/` prefix fix, PR #412) which fixes image-reference handling in upgrade tests.
- **`common` tests (also SSH-only) are currently passing** in every CI run, confirming SSH infrastructure and the bootc test environment are healthy. Lifecycle tests are structurally equivalent.

**Changes:**
- Uncomment the `run-upgrade-test` job
- Add it back to `promote-to-testing`'s `needs:` array
- Restore gate condition: `needs.run-upgrade-test.result == 'success'`
- Remove the TODO comment and disabled-state block comment

## Checklist

- [x] PR title follows Conventional Commits (`ci:`)
- [x] Workflow YAML is valid (no actionlint config changes needed; `@v1` tag is exempt per pre-commit rule)
- [x] No SHA pinning regression (uses `projectbluefin/actions/.github/workflows/upgrade-test.yml@v1` per repo convention)

## AI attribution

Assisted-by: Claude Sonnet 4.6 via GitHub Copilot
Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

PR #972, opened 2026-08-04
ci: restore lifecycle upgrade test promotion gate

### LINKED ISSUE #400: ci: restore upgrade-test (lifecycle) as a promotion gate once GNOME 50 AT-SPI is stable
## Context

`post-testing-e2e.yml` runs a lifecycle (upgrade & rollback) suite via
`projectbluefin/actions/.github/workflows/upgrade-test.yml`. The lifecycle suite
is failing under GNOME 50 AT-SPI changes (see run
[27075323936](https://github.com/projectbluefin/bluefin/actions/runs/27075323936),
job: `GNOME 50 — lifecycle`).

In PR #399 the `run-upgrade-test` job was decoupled from the `promote-to-testing`
and `report-failure` gates so that lifecycle test failures don't block the
continuous promotion of `:testing` or the weekly promotion workflow. The job
still runs and uploads results artifacts.

## Steps to restore

1. Investigate and fix the lifecycle test failures in `projectbluefin/testsuite`
   (GNOME 50 AT-SPI changes documented in the bluefin-ci skill and testsuite ops.md).
2. Once lifecycle scenarios pass consistently in CI, restore `run-upgrade-test` to
   the `promote-to-testing` and `report-failure` `needs:` arrays in
   `.github/workflows/post-testing-e2e.yml`.
3. Also restore the gate condition:
   ```yaml
   if: >-
     needs.run-e2e.result == 'success' &&
     needs.run-upgrade-test.result == 'success'
   ```
4. Remove the TODO comment from `run-upgrade-test`.

## Labels
area/testing, kind/bug

**Bluefin · issue pipeline**

```text
  ▶  triage     needs kind/ + area/ then /approve
  ·  discussing —
  ·  queued     —
  ·  claimed    —
  ·  done       —
```

**area:** area/testing &nbsp; **priority:** —  
**maintainer:** set kind/ + area/ labels, then comment /approve  
**reporter:** answer questions; add repro steps if a bug

castrojo:
## Summary

- Re-enable the lifecycle `run-upgrade-test` job in `post-testing-e2e.yml`.
- Require lifecycle success alongside smoke/common E2E before promoting to `:testing`.
- Closes #400.

## Validation

- `git diff --check`
- Python workflow assertions

`pre-commit` and `just` were unavailable in the contributor environment.

### DIFF
--- a/.github/workflows/post-testing-e2e.yml
+++ b/.github/workflows/post-testing-e2e.yml
@@ -54,29 +54,23 @@ jobs:
       image: ${{ needs.e2e.outputs.image }}
       suites: smoke,common
 
-  # run-upgrade-test: DISABLED — TODO(#400)
-  # lifecycle suite fails under GNOME 50 AT-SPI changes, causing post-testing-e2e
-  # to mark as 'failure' and blocking promotion.
-  # Disabled until testsuite lifecycle scenarios are stable on GNOME 50.
-  # To restore: uncomment the job below.
-  #
-  # run-upgrade-test:
-  #   needs: [e2e]
-  #   permissions:
-  #     packages: write
-  #     contents: read
-  #   uses: projectbluefin/actions/.github/workflows/upgrade-test.yml@v1
-  #   with:
-  #     image: ${{ needs.e2e.outputs.image }}
-  #     suites: lifecycle
+  run-upgrade-test:
+    needs: [e2e]
+    permissions:
+      packages: write
+      contents: read
+    uses: projectbluefin/actions/.github/workflows/upgrade-test.yml@v1
+    with:
+      image: ${{ needs.e2e.outputs.image }}
+      suites: lifecycle
 
   promote-to-testing:
-    # Promote all tested flavors to :testing only after e2e passes.
-    # This ensures :testing always points to a digest that has passed gate e2e.
-    # run-upgrade-test removed from needs — lifecycle suite is non-blocking (see TODO above).
-    needs: [e2e, run-e2e]
+    # Promote all tested flavors to :testing only after e2e and lifecycle pass.
+    # This ensures :testing always points to a digest that has passed all gate suites.
+    needs: [e2e, run-e2e, run-upgrade-test]
     if: >-
       needs.run-e2e.result == 'success' &&
+      needs.run-upgrade-test.result == 'success' &&
       github.event.workflow_run.head_branch == 'main'
     runs-on: ubuntu-latest
     timeout-minutes: 15


castrojo:
Closing this in favor of consolidating on #913 (byte-identical diff), and closing that one too — see the shared rationale below. Not a rubber-stamp: I went and looked for evidence that lifecycle recovered, and it hasn't.

## Why this can't merge yet

**1. #400 is still open and still blocked.** Last activity 2026-07-28. Its own comment states the resolution plainly: *"No in-repo code change can unblock this — it requires the testsuite fix first."* Nothing in this diff is a testsuite fix.

**2. There is no green lifecycle run anywhere to point at.** `projectbluefin/actions/.github/workflows/upgrade-test.yml` — the exact reusable workflow this job calls — has four runs in its entire history:

| date | conclusion |
|---|---|
| 2026-07-19 | cancelled ×3 |
| 2026-06-03 | **failure** |

Zero successes. And no current Bluefin workflow exercises `lifecycle` at all — `nightly.yml` runs `smoke,common,vanilla-gnome`, `e2e-dispatch.yml` runs `smoke,developer,vanilla-gnome`. So the suite has had no signal for ~2 months. hanthor's ask on #913 was "a link to a green upgrade-test / lifecycle run against a recent image" — that link does not exist.

**3. `:testing` promotion is *already* blocked, by a different bug.** Every `Post-Testing E2E` run for the last several days ends with `promote-to-testing: skipped` because `run-e2e / GNOME 50 — smoke-a` fails:

- [30931237054](https://github.com/projectbluefin/bluefin/actions/runs/30931237054) — smoke-a failure, promote-to-testing skipped
- [30931128732](https://github.com/projectbluefin/bluefin/actions/runs/30931128732) — same
- [30833806016](https://github.com/projectbluefin/bluefin/actions/runs/30833806016) — same
- [30832171120](https://github.com/projectbluefin/bluefin/actions/runs/30832171120) — same
- [30758889324](https://github.com/projectbluefin/bluefin/actions/runs/30758889324) — same

Adding a second, known-unstable gate on top of an already-red pipeline adds no protection and buries the real regression.

**4. Wrong base branch.** Both PRs target `main`. Per `AGENTS.md` / `docs/contributing.md`, feature work targets `testing`; `main` is fed by `promote-testing-to-main.yml`. Any replacement PR must be based on `testing`.

## The `needs:`/`if:` semantics, verified against the file

`promote-to-testing` currently has `needs: [e2e, run-e2e]` with no `always()`. A failed upstream job therefore **skips** the whole job — which is exactly what the run data above shows. Adding `run-upgrade-test` to `needs` plus `needs.run-upgrade-test.result == 'success'` to `if:` makes that strictly stricter, with no partial-credit path. hanthor's read of this was correct.

## Why I did not take the "non-blocking informational job" middle path

I tested it rather than guessing. Leaving `run-upgrade-test` out of `promote-to-testing`'s `needs:` genuinely prevents it from blocking promotion — but you cannot mark it non-blocking at the workflow level, so the run would report `failure` forever. `actionlint` confirms:

```
when a reusable workflow is called with "uses", "continue-on-error" is not available.
only following keys are allowed: "name", "uses", "with", "secrets", "needs", "if", and "permissions"
```

So the options are a permanently-red workflow plus ~1h of extra QEMU time on every push build, against a suite with no green baseline to diff against — noise, not signal. `nightly.yml` is also red on every run right now, so parking it there doesn't help either. Once smoke-a is fixed and the testsuite side of #400 lands, an informational run becomes worth doing; today it isn't.

## What happens next

The intent here is right and I don't want it lost — a commented-out gate is not a gate. I've recorded these findings on #400 so restoring the gate stays tracked with concrete unblock criteria. Please re-open a PR against `testing` once #400's testsuite work has landed and there's a green lifecycle run to cite.

PR #1016, opened 2026-08-07
fix(build): give Stage 2 and the ISO stage narrow ctx mounts

### LINKED ISSUE #996: build: narrow ctx mounts for Stage 2 and the ISO stage
Follow-up from #964, deliberately left out of scope there.

## What
#964 added a narrow `ctx-build` stage (`build_files` + `image-versions.yml` only) so the package-install stage stops being invalidated by unrelated edits. While verifying that fix, testing confirmed the same class of problem remains in the other stages:

Stage 2 and the ISO stage still mount from the wide `ctx`, which carries **both** `build_files/` and `system_files/`. Buildah folds the mounted stage's image ID into the consuming `RUN`'s cache key, so any edit to either directory invalidates both stages together.

## Why it matters
The Containerfile documents a "saves 20-80 min" benefit from granular layer caching. That benefit is only partially realised while these stages share a wide context — a `system_files`-only change still pays for work it cannot have affected.

## Fix
Give each stage a context stage narrow enough to cover only what it actually reads, following the `ctx-build` pattern established in #964.

## Verification
The approach used in #964: build twice against isolated `--root` stores with a local registry cache, change only one input directory, and confirm the unaffected stage reports a cache hit.

Automatable: yes

castrojo:
Closes #996. Follow-up to #964, which established the `ctx-build` pattern and deliberately left these two stages out of scope.

## Problem

Both remaining consumers mounted from the wide `ctx`, which carried `build_files/` **and** `system_files/`. Since buildah folds the mounted stage's image ID into the consuming `RUN`'s cache key (the mechanism #964 verified empirically), an edit anywhere in either directory invalidated Stage 2 and the ISO layer together.

## Change

Two edits, no behaviour change:

- **`ctx` narrowed to what Stage 2 actually reads** — `system_files/`, `build_files/shared/`, and the four `build_files/base` scripts the stage invokes (`00-image-info.sh`, `17-cleanup.sh`, `19-initramfs.sh`, `20-tests.sh`). The package-install scripts (`03`/`04`/`05`) are gone from it, so editing them can no longer invalidate Stage 2.
- **New `ctx-iso` scratch stage** — the ISO `RUN` reads exactly two files (`21-container-native-iso.sh` and `shared/utils/ghcurl`), so it gets a context holding only those.

The `--mount` lines and every `/ctx/...` path inside the `RUN` payloads are unchanged; only the `from=` stage and what each context copies changed.

## Why `ctx` keeps `build_files/shared` whole

`17-cleanup.sh` does `source /ctx/build_files/shared/disable-repos.sh` (line 49), and Stage 2 invokes `finalize-gnome-extensions.sh`, `validate-repos.sh`, and `clean-stage.sh` from that directory. Enumerating those files individually would add four more `COPY` lines that must be kept in sync with the payload for no cache benefit — the whole directory is 5 scripts plus `utils/`.

## Verification

I could not run the empirical two-build cache check the issue asks for. **This runtime has no container tooling** — `command -v podman buildah` both return nothing, and there is no package manager to install them. That check needs a maintainer or CI.

What I did verify:

- **Every bind mount resolves against its context stage.** Script-checked all 12 `--mount=type=bind` lines against the `COPY` destinations of the three `scratch` stages: all resolve. Every `COPY` source also exists in the working tree.
- **No unmounted `/ctx/...` reference.** Script-extracted every `/ctx/…` path from each `RUN` payload and confirmed it falls under one of that `RUN`'s own mount targets. A missing script would fail the build loudly at that `RUN`, not silently.
- **`build_files/base/18-workarounds.sh` is intentionally absent** — `grep -n 18-workarounds Containerfile` finds nothing; it is only referenced by the unused `build_files/shared/build.sh`. Not a regression from this change.
- `just check` passes. `python3 .github/scripts/validate-docs.py` passes (13 skills, 39 files).
- **Not run:** `hadolint` and `pre-commit` are not installed here. `shellcheck` is scoped to `^(build_files|system_files)/.*\.sh$` and no shell script changed.

## Risk

If a future edit adds a `build_files/base` script to Stage 2 without adding it to `ctx`, the build fails at that `RUN` with a missing path — noisy, not silent. I noted that in the stage comment.

Assisted-by: Claude via GitHub Copilot
Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -20,9 +20,25 @@ FROM scratch AS ctx-build
 COPY /build_files /build_files
 COPY /image-versions.yml /image-versions.yml
 
+# ISO context. Same reasoning as `ctx-build`: the ISO RUN reads exactly two
+# files, so giving it its own scratch stage keeps an edit to
+# 21-container-native-iso.sh from invalidating Stage 2, and keeps an edit to
+# any other build_files script from invalidating the ISO layer.
+FROM scratch AS ctx-iso
+COPY /build_files/base/21-container-native-iso.sh /build_files/base/21-container-native-iso.sh
+COPY /build_files/shared/utils/ghcurl /build_files/shared/utils/ghcurl
+
+# Overlay context for extension-builder and Stage 2. Carries system_files plus
+# only the build_files that run after Stage 1; the package-install scripts
+# (03/04/05) are deliberately absent so editing them cannot invalidate Stage 2.
+# A script added to Stage 2 must be added here too, or the build fails loudly.
 FROM scratch AS ctx
 COPY /system_files /system_files
-COPY /build_files /build_files
+COPY /build_files/shared /build_files/shared
+COPY /build_files/base/00-image-info.sh /build_files/base/00-image-info.sh
+COPY /build_files/base/17-cleanup.sh /build_files/base/17-cleanup.sh
+COPY /build_files/base/19-initramfs.sh /build_files/base/19-initramfs.sh
+COPY /build_files/base/20-tests.sh /build_files/base/20-tests.sh
 COPY --from=common /system_files/shared /system_files/shared
 COPY --from=common /system_files/bluefin /system_files/shared
 COPY --from=brew /system_files /system_files/shared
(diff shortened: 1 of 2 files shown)


bketelsen:
I found one remaining coupling in this version: `COPY /build_files/shared /build_files/shared` includes Stage 1-only inputs (`copr-helpers.sh`, `package-lib.sh`, and `read-packages`), so edits to those files still change the Stage 2 mount context. I opened #1020 on top of this commit with exact shared-file copies, regression coverage, and a successful full image build. The incremental correction is commit `c4cbc9a` if you prefer to incorporate it here.

PR #1082, opened 2026-08-10
docs(security): correct CVE-2026-33186 root cause — podman is a second carrier

Danathar:

I picked up #919 expecting to confirm the existing triage and move on. The conclusion that there is no in-repo fix is correct, but the reason recorded in our security docs is not: it names `buildah` as the only carrier of the vulnerable grpc and tells the reader that an F44 buildah update will clear the finding. It won't — `podman` ships the same vulnerable module version. Anyone tracking this per the current doc would be waiting on an update that cannot close the alert.

---

## The finding

Verified against the exact versions Fedora 44 ships **today (2026-08-10)**, RPM versions from `mdapi.fedoraproject.org` and grpc pins from each project's `go.mod` at the shipped tag:

| F44 package | grpc pinned | Status |
|---|---|---|
| `buildah-2:1.43.2-1.fc44` | `v1.72.2` | vulnerable |
| **`podman-5.8.4-1.fc44`** | **`v1.72.2`** | **vulnerable** |
| `skopeo-1.22.2-2.fc44` | `v1.79.3` | already fixed |

`skopeo` is in the table on purpose: the same container-tools stack already carries a fixed build, so "container-tools is vulnerable" is too coarse a statement. The exposure is per-binary, and fixing one carrier does not clear the scan while the other stands.

**The podman side is also the harder one:**

| Project | First fixed release | grpc after fix |
|---|---|---|
| `buildah` | `v1.44.0` | `v1.81.1` |
| `podman` | `v6.0.2` | `v1.81.1` |

buildah's fix is a minor bump Fedora could reasonably carry. podman's is not fixed until **v6.0.2** — the latest `5.8.x` release, `v5.8.5`, still pins `grpc v1.72.2`. There is no podman 5.x release that resolves this, so clearing it on F44 would need Fedora either to rebase podman 5.8.x → 6.x inside a stable release (unlikely) or to carry a backported grpc bump.

That inverts the doc's advice: the **Fedora 45 migration is the primary path**, not the fallback it was written as.

## Changes

Docs only — two files, no build or scan behavior touched.

- `docs/skills/security/references/cve-2026-33186-grpc-buildah.md` — corrected root cause (two carriers, with the skopeo contrast), corrected "where the fixes land" including why podman is the blocker, and reworked the Action section accordingly.
- `docs/skills/security/SKILL.md` — link text now says "buildah and podman". **Filename left alone** so the existing link and any external references keep resolving, even though it now slightly under-describes the content.

Also added a **re-check section**. This CVE has now been triaged from scratch at least four times (#870, two rounds on this issue, and this one), each time by redoing the same lookups by hand. Both facts are now one command each.

One note on that: I wrote the mdapi recipe with `grep` first, ran it, and it printed several hundred lines — an mdapi response embeds a `"version"` field for *every bundled Go module*, and the RPM's own version is easy to mistake for one of them. The committed version parses with `jq` and carries its expected output inline. Both commands in the doc were executed as written before committing.

## What I did not do

- **No suppression, no `.trivyignore`, no gate weakening.** The "do not suppress" guidance is unchanged. Accepting a critical is a Security-gate decision, not mine.
- **No package removal or pin.** There is still no fixed F44 RPM for either carrier to pin or override to.
- **No change to the auto-issue behavior.** Worth correcting one point from earlier triage while I'm here: the dedup is not filing an issue per build. Searching all issues titled "critical CVE detected" returns four in about two months (#530, #638, #870, #919), so this is roughly monthly, not per-build.

## Verification

- [x] F44 versions via `mdapi.fedoraproject.org` for buildah/podman/skopeo
- [x] grpc pins read from `go.mod` at `buildah@v1.43.2/v1.44.0`, `podman@v5.8.4/v5.8.5/v6.0.2`, `skopeo@v1.22.2`
- [x] Both commands added to the doc executed as written; the `jq` one's output is pasted into the doc
- [ ] Trivy re-scan — not run; no container toolchain in this runtime
- [ ] Which carrier a specific alert instance points at — reading code-scanning alert locations returned HTTP 403 for me. It does not change the conclusion, since both carriers must be fixed regardless, and the doc says so explicitly rather than implying I checked.

## Checklist

- [x] Docs updated (this is the docs update)

Refs #919. **Does not close it** — the CVE is still present and still has no in-repo fix. This makes the tracking note accurate about *what* to track.

### DIFF
--- a/docs/skills/security/SKILL.md
+++ b/docs/skills/security/SKILL.md
@@ -50,7 +50,7 @@ do not invent a replacement key or trust path.
 
 - [COPR isolation invariant](references/copr-isolation.md)
 - [signing and verification](references/signing.md)
-- [CVE-2026-33186 grpc in buildah](references/cve-2026-33186-grpc-buildah.md)
+- [CVE-2026-33186 grpc in buildah and podman](references/cve-2026-33186-grpc-buildah.md)
 
 ## When to Use
 
--- a/docs/skills/security/references/cve-2026-33186-grpc-buildah.md
+++ b/docs/skills/security/references/cve-2026-33186-grpc-buildah.md
@@ -61,6 +61,15 @@ own; re-scan and confirm rather than assuming. Do not weaken the vulnerability
 scan gate or add an exception; this finding is a tracking note, not an approval
 to suppress the scan.
 
+## A new issue number is not a new problem
+
+`bootc-build/scan-image` files a fresh `fix(security): critical CVE detected …`
+issue for non-PR builds, and its duplicate check keys on the ref, so a `main`
+build does not match an existing `testing` issue. This CVE has been filed at
+least four times (#530, #638, #870, #919) — roughly monthly, not per build.
+Before triaging one of these as new, compare the CVE ID against this file. The
+dedup logic lives in `projectbluefin/actions`, not here.
+
 ## Re-checking this (do not re-derive by hand)
 
 This finding has been re-triaged several times, each time by repeating the same
--- a/docs/skills/security/references/cve-2026-33186-grpc-buildah.md
+++ b/docs/skills/security/references/cve-2026-33186-grpc-buildah.md
@@ -1,50 +1,109 @@
-# CVE-2026-33186 (google.golang.org/grpc) in the buildah binary
+# CVE-2026-33186 (google.golang.org/grpc) in the Fedora container-tools stack
 
 Severity: **Critical** (CVSS ≥ 9.0, per the vulnerability scan gate).
 
 ## Finding
 
-CVE-2026-33186 affects `google.golang.org/grpc v1.72.2`, which is bundled in the
-`buildah` binary shipped by the Fedora container-tools stack. There is **no
-self-contained fix** that this repository can apply on the Fedora 44 target.
+CVE-2026-33186 affects `google.golang.org/grpc` before `v1.79.3`, which is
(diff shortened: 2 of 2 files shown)


mrbobbytables (approved):
test approval

mrbobbytables (approved):
test approval2

mrbobbytables (commented):
(Correcting my two prior placeholder reviews above, which were wrapper-debugging artifacts — please disregard the 'test approval'/'test approval2' bodies.)

Re-verified 2026-08-12 before reviewing substantively:
- Fedora 44 stable is still `buildah-1.43.2-1.fc44` / `podman-5.8.4-1.fc44` (mdapi, live query).
- Bodhi shows no new update submitted for either package on F44 since 2026-06-19 (buildah) / 2026-07-03 (podman) — nothing in testing/pending that bumps grpc.
- `google.golang.org/grpc` remains pinned at the vulnerable `v1.72.2` in both projects; the fix requires `buildah >= v1.44.0` and `podman >= v6.0.2`, neither available on F44.
- This diff correctly identifies podman as a second carrier (skopeo already fixed at v1.79.3), fixes the SKILL.md doc link, and reframes the F45 migration as the primary path. Matches my independent verification.

No objections to the content. Recommend merging so the docs stop pointing at the incomplete (buildah-only) root cause carried since the original #870 triage.

Issue #939, opened 2026-08-01
ci: grant MergeRaptor app Checks write permission for Lab check


castrojo:
What: `.github/workflows/lab-check.yml` cannot mint its existing MergeRaptor installation token because the installation does not grant the requested `checks: write` permission. The workflow has failed repeatedly before it can create or update the `testing-lab / bluefin` check.

Evidence: the workflow job annotation is `The permissions requested are not granted to this installation.` Organization secrets `MERGERAPTOR_APP_ID` and `MERGERAPTOR_PRIVATE_KEY` already exist and are available to all repositories.

Fix: grant the existing MergeRaptor GitHub App installation **Checks: write** permission, then dispatch or wait for a `lab-check` repository event and confirm it updates the check run. Do not create a PAT or rotate/create credentials.

Automatable: no; GitHub App permission administration is required.

mrbobbytables:
Investigated — this is not fixable with a code change. `.github/workflows/lab-check.yml` already requests the correct scope: line 24 sets `permission-checks: write` when minting the MergeRaptor installation token via `actions/create-github-app-token`. The workflow's token-request and check-run update logic are both correct as written.

The failure (`The permissions requested are not granted to this installation`) means the **MergeRaptor GitHub App installation** itself has not been granted `Checks: write` at the org/installation level. That can only be changed by an org owner, via GitHub UI: Organization settings → GitHub Apps → MergeRaptor → Permissions (approve the pending permission request), or Installed GitHub Apps → MergeRaptor → Configure → grant Checks: write.

No repository change, secret, or credential is being proposed — per repo policy I won't create/rotate any app credentials, and there is no code fix available here. @castrojo — could you or another org owner approve/grant `Checks: write` to the MergeRaptor app installation? Once granted, dispatching a `lab-check` repository event (or the next `track-common` QA run) should let it update the `testing-lab / bluefin` check successfully. Happy to help verify once it's granted.

---
🐝 **Hive Agent**: `contributor` | **SHA:** unknown

hanthor:
Confirmed the prior finding: `lab-check.yml` is correct as written (requests `permission-checks: write` via `actions/create-github-app-token`), and the failure is the MergeRaptor app installation lacking **Checks: write** — an org-owner-only fix in the GitHub UI (the repo reference doc `docs/skills/ci/references/mergeraptor-checks.md` already documents this).

[...]

PR #1079, opened 2026-08-09
ci(lab-check): point the Checks-permission fix at the right screen

Danathar:
Refs #939. **This does not grant the permission** — that is org-owner UI work, exactly as the issue says. It fixes the instructions for doing it, which are currently one step short.

## The gap

`lab-check.yml`'s error message and `docs/skills/ci/references/mergeraptor-checks.md` both send an org owner to the **installation** page:

> an org owner must grant it (Organization settings → GitHub Apps → MergeRaptor → Permissions)

But the measurement posted on #939 shows `checks` is absent from the App's **own declared permissions**, not merely ungranted on the installation:

```console
$ gh api /apps/mergeraptor --jq '.permissions'
{"contents":"write","metadata":"read","pull_requests":"write","workflows":"write"}
```

An App can only be granted what it declares. So an owner who follows the current instructions arrives at the installation page and finds **nothing to approve** — as @castrojo put it, *"re-installing or re-approving cannot help; the App definition has to request it first."* The declaration lives on a different screen: Developer settings → GitHub Apps → MergeRaptor → Permissions & events → Repository permissions.

That is a plausible reason this has stayed broken. I re-measured today and **Lab check is still failing 12 out of 12 runs**:

```console
$ gh run list --repo projectbluefin/bluefin --workflow "Lab check" --limit 12 \
    --json conclusion --jq '[.[] | .conclusion] | group_by(.) | map({(.[0]): length}) | add'
{"failure":12}
```

## What changed

- **Error message** now names both steps, gated on the diagnosis so nobody does step 1 when only step 2 is needed.
- **Reference doc** gains the App-level query (`gh api /apps/mergeraptor`) alongside the existing installation-level one, plus a table mapping the two readings to the required action — absent/absent means both steps, present/absent means approve only, present/write means look elsewhere.
- Notes that both commands need `admin:org`, and a contributor token returns `404` — which is *not* evidence the grant is missing. I hit exactly that while verifying, so it seemed worth writing down.
- Writes the **"never add a fallback that lets the job pass without reporting"** rule into the reference, so the next person does not relitigate it.

## Explicitly unchanged

@castrojo's note on the issue was that no repository change is warranted and that a fallback skipping the report would turn a loud gap into a silent one. **I agree and did neither.** The token step still requests exactly `permission-checks: write` and `permission-contents: read`; the diagnostic step still `exit 1`s; no fallback was added; no credential is created, rotated, or proposed. The only behavioral surface touched is the text inside an `echo`.

[...]

PR #1083, opened 2026-08-10
ci(lab-check): name both org-owner steps in the Checks-grant error

Danathar:

The real fix for #939 is org-owner UI work and nothing in a PR can do it — I'm not trying to. But the error message we print when it fails sends the reader to the wrong settings page. It names the *approval* step, and @castrojo's own investigation on the issue showed approval is not the step that's missing: `checks` is absent from the MergeRaptor app definition, so there is no request for an owner to approve. Someone following our message lands on a page with nothing to click.

---

## The gap

A GitHub App permission exists at two levels, and the token-mint error is identical either way:

| Level | Meaning | Where it changes |
|---|---|---|
| **Declared** on the app | the app *asks* for it | Org settings → Developer settings → GitHub Apps → MergeRaptor → Permissions & events |
| **Granted** on the installation | an owner *approved* what was asked | Org settings → GitHub Apps → MergeRaptor |

An installation can never be granted a permission the app does not declare. From #939:

> `checks` is not merely ungranted on the installation — it is not among the App's declared permissions at all. So re-installing or re-approving cannot help; the App definition has to request it first.

Both the diagnostic from #1058 and the reference doc from #1010 point only at the second level. That was accurate as far as it went, and it is exactly the half that cannot currently be acted on.

## Changes

**`.github/workflows/lab-check.yml`** — one line, the `::error::` text. It now names both steps in order (declare on the app → approve on the installation), says in one clause why step 2 alone is a no-op, and points at the two commands that distinguish declared from granted so the reader can tell which step they still owe.

**`docs/skills/ci/references/mergeraptor-checks.md`** — the same distinction as a table, a which-level-am-I-missing recipe with the three possible readings, and the fix as two ordered steps.

I also documented something that cost me time while verifying: the token step reports **conclusion `success` while its `outcome` is `failure`**, because of `continue-on-error`. Reading a failed run, "Get MergeRaptor token: success" followed by a failing diagnostic looks like a second bug. It isn't — the `if` keys off `outcome`, correctly — but it's worth a sentence so the next person doesn't chase it. I did chase it, and confirmed from the run log that the mint genuinely 422s.

## Explicitly not doing

@castrojo's note that *"adding a fallback that skips the report would convert a loud, accurate gap into a silent one"* is right, and this PR does not do that. The step fires on the same condition, still `exit 1`s, and the gate is unchanged — the failure stays exactly as loud. The only change is that the text now describes a fix that can actually be carried out. On the "no repository change is warranted" reading: I'd offer that the message being *complete* is the one repo-side thing still worth correcting, since the current text can't be followed to completion. Happy to drop this if you'd rather keep it frozen.

## Verification

- [x] Extracted the diagnostic's `run:` block and **executed it** — emits exactly one `::error::` annotation and exits 1. Worth doing: my first draft embedded the `--jq` filters inline and the nested single-quote escaping was wrong; the committed version keeps the full commands in the doc instead.
- [x] Confirmed the current failure mode from run `31343742552` (2026-08-10): mint returns HTTP 422 `The permissions requested are not granted to this installation`, diagnostic fires, `Create or update lab check` is skipped, job fails.
- [x] Re-confirmed the state is unchanged today — last 12 Lab check runs all `failure`, most recent `2026-08-10T00:10`.
- [x] Diff is one line of workflow plus docs; no trailing whitespace or tabs; block-scalar indentation unchanged.
- [ ] `actionlint` — not run locally (unavailable here, along with any YAML parser); CI covers it. The changed line lives inside a `run: |` block scalar at unchanged indentation, so it cannot alter the document structure.
- [ ] Could not read app/installation permissions myself — both endpoints 404 without org-owner scope. That limitation is now written into the doc, since a 404 is not a diagnosis.

## Checklist

- [x] No hardcoded secrets or credentials; no PAT, no credential creation or rotation
- [x] No change to the gate's pass/fail behavior

Refs #939. **Does not close it** — the org-owner grant is still required, and the check will stay red until then.

### DIFF
--- a/.github/workflows/lab-check.yml
+++ b/.github/workflows/lab-check.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Fail with actionable message if token minting was denied
         if: steps.app-token.outcome == 'failure' || steps.app-token.outputs.token == ''
         run: |
-          echo "::error::MergeRaptor token mint failed — the GitHub App installation almost certainly lacks the requested 'checks: write' permission. This cannot be fixed from the repository; an org owner must grant it (Organization settings → GitHub Apps → MergeRaptor → Permissions). Verify current grants with: gh api orgs/projectbluefin/installations --jq '.installations[] | select(.app_slug == \"mergeraptor\") | .permissions' — see docs/skills/ci/references/mergeraptor-checks.md."
+          echo "::error::MergeRaptor cannot mint a 'checks: write' token. Not fixable from this repository — it needs TWO org-owner steps in the GitHub UI, in this order: (1) DECLARE it on the app: Organization settings -> Developer settings -> GitHub Apps -> MergeRaptor -> Permissions & events -> Repository permissions -> Checks: Read and write. (2) APPROVE the request that raises on the installation: Organization settings -> GitHub Apps -> MergeRaptor. Step 2 alone does nothing while 'checks' is absent from the app definition, because there is no request to approve. To see which step you are on, compare what the app DECLARES (gh api /apps/mergeraptor) against what the installation was GRANTED (gh api orgs/projectbluefin/installations); both commands are spelled out in docs/skills/ci/references/mergeraptor-checks.md."
           exit 1
 
       - name: Create or update lab check
--- a/docs/skills/ci/references/mergeraptor-checks.md
+++ b/docs/skills/ci/references/mergeraptor-checks.md
@@ -5,12 +5,62 @@ installation already holds; anything else fails the run with `The permissions
 requested are not granted to this installation.` The MergeRaptor installation
 on `projectbluefin` therefore needs **Checks: write** for `lab-check.yml`, in
 addition to the contents/pull-requests/workflows grants `track-common.yml`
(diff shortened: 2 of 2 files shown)


Danathar:
Closing as a duplicate of #1079 — my mistake. I opened this without checking for existing PRs on #939, and #1079 (opened ~4 hours earlier, same fork) makes the same two-step correction to the same two files.

#1079 is the better version: its diagnosis table spells out all three declared/granted combinations and what each one means, and it keeps the no-PAT / no-silent-fallback rules together in one section.

The one thing this PR had that #1079 didn't, I've pushed to #1079 instead (`docs(ci): explain the success/failure split when reading a failed lab-check`): a failed run shows `Get MergeRaptor token: success` right above the failing diagnostic, which reads like a second bug. It's just `continue-on-error` — conclusion `success`, outcome `failure`, and the diagnostic keys off `outcome`. Verified against run `31343742552`, where the mint logs a 422 while the step still shows as success.

Nothing is lost by closing this. Review #1079.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `c078f7e`

Issue #989, opened 2026-08-07
ci: `:testing` promotion frozen — `run-e2e / GNOME 50 — smoke-a` failing since 2026-06-25 (firefox.feature AT-SPI)


castrojo:
## Summary

`promote-to-testing` in `.github/workflows/post-testing-e2e.yml` has been `skipped`
on essentially every run since **2026-06-22**, because the matrix leg
`run-e2e / smoke,common / GNOME 50 — smoke-a` fails. The failure is **not**
6 days old — it is ~46 days old and 100% reproducible.

Filed because this has had no tracking issue.

## 1. Blast radius

`gh run list --workflow post-testing-e2e.yml --limit 300` (back to 2026-06-16):

| conclusion | runs |
|---|---|
| failure | 209 |
| skipped | 89 |
| **success** | **2** |

The only two successes were `27922429853` and `27922740150`, both on
**2026-06-22**. Every run since has been `failure` (or `skipped` when the
upstream build didn't qualify). `promote-to-testing` has therefore been dead
for ~46 days.

Registry state (`skopeo inspect docker://ghcr.io/projectbluefin/bluefin:testing`):

```
Digest:  sha256:bf615b200faefc44b50232ecc8a3eb21490e88dd9316b528b27f54472122611d
Created: 2026-08-04T16:32:43Z
Tags on this version: testing, testing-20260804, testing-44.20260804
```

So `:testing` is ~2.5 days stale as of 2026-08-07.

### ⚠️ Secondary finding — the gate is being bypassed

`sha256:bf615b20…` is **the exact digest that run
[30931237054](https://github.com/projectbluefin/bluefin/actions/runs/30931237054)
tested and FAILED** (`IMAGE: ghcr.io/projectbluefin/bluefin@sha256:bf615b20…`),
and in that run `promote-to-testing` was `skipped`.

`build-image-testing.yml` passes `publish_stream_tag: "false"`, and
`projectbluefin/actions/.github/workflows/reusable-build.yml@v1` (`compute-push-tags`,
~L375–392) is supposed to exclude the bare stream tag for exactly this reason.
Yet the bare `testing` tag landed on that digest anyway. No other workflow in
this repo writes `:testing` (`promote-testing-to-main.yml` does not; `nightly.yml`
and `pr-validation.yml` only read it).

**This needs investigation in `projectbluefin/actions`:** either the
`compute-push-tags` output is not honored by the push step (e.g. `just tag-images`
[...]

PR #1086, opened 2026-08-10
docs(ci): record that ping has no cap_net_raw on Fedora by design

Danathar:

One of the four threads blocking `:testing` promotion is the smoke-b composefs assertion failing on `/usr/bin/ping`. It has been triaged twice now as *possibly* a bluefin image defect — most recently on #989 as "an image-content finding on bluefin (or a mis-scoped re-activation in testsuite)". It is the second one, and the evidence settles it conclusively. This records that so it stops being re-litigated, and so nobody "fixes" it in the image.

---

## The evidence

Fedora's `iputils` ships `ping` with **no file capabilities at all**. From `iputils.spec` on the `f44` branch — the exact package this image gets:

```spec
%attr(0755,root,root) %caps(cap_net_raw=p) %{_bindir}/clockdiff
%attr(0755,root,root) %caps(cap_net_raw=p) %{_bindir}/arping
%attr(0755,root,root) %{_bindir}/ping
```

`clockdiff` and `arping` get `cap_net_raw=p`. `ping` is a bare `%attr` — it works through unprivileged ICMP sockets via `net.ipv4.ping_group_range`, which systemd sets by default. **There is no xattr on `ping` to preserve or strip**, so:

```
["/usr/bin/ping: expected 'cap_net_raw' in ''"]
```

cannot pass on any Fedora-based image regardless of what composefs does.

**The same failure proves composefs is healthy.** The other two binaries in that assertion genuinely do carry capabilities — `shadow-utils` sets `cap_setuid=ep` on `newuidmap` and `cap_setgid=ep` on `newgidmap` — and both now pass, where earlier runs failed all three. Capability preservation is working. The expectation list is what's wrong.

That also explains the narrowing noted on #929: the assertion went from naming three binaries to naming only `ping`. That wasn't partial progress on a composefs bug — it was the two real capability-carrying binaries starting to pass, leaving behind the one that never had a capability.

## Changes

One section in `docs/skills/ci/references/failure-modes.md`, next to the existing "The `oras` screenshot error is not the failure" entry — same genre of "this red line is not your bug".

It includes an explicit **do not** : don't add `setcap cap_net_raw` to `ping` in the image to turn the test green. That diverges from Fedora for no functional gain, since `ping` already works. And a constructive alternative for the testsuite side: assert on `arping` or `clockdiff` instead, which really do carry `cap_net_raw=p` — that would make the scenario an actual composefs test rather than one that can only ever fail.

## Scope — what this does and does not do

This does **not** unblock the gate. `firefox.feature` (thread 1) is still the deterministic blocker and needs live-session AT-SPI instrumentation in testsuite; nothing in this repo addresses it.

What it does is correctly reassign thread 2 away from bluefin, with proof, so the remaining work is scoped to testsuite.

I also want to flag the circuit-breaker proposal in the last issue comment. Option A adds a `force_promote` dispatch input so `promote-to-testing` can run when `run-e2e` failed. That is the outcome this issue's own **"Explicitly NOT the fix"** section rules out — "Do not add `always()`, `continue-on-error`, or drop `run-e2e` from `promote-to-testing`'s `needs:`. That promotes an unverified digest and is the failure mode a sibling PR was already closed for." A manual bypass reaches the same end by a different mechanism, and it is especially risky here given §1 of the issue: `:testing` already points at an e2e-**failing** digest, so the gate is leaking without anyone adding a bypass. I have not implemented it.

## Testing

- [x] `python3 .github/scripts/validate-docs.py` → `documentation ok: 13 skills, 41 Markdown files`
- [x] Capability claims read from Fedora dist-git `f44` branches of `iputils` and `shadow-utils`, not from memory
- [x] No trailing whitespace or tabs
- [ ] Not verified against a running image — no container/VM toolchain here. The claim is about what the RPM ships, which the spec settles; a `getcap /usr/bin/ping` on a booted image would confirm it returns empty, as expected.

## Checklist

- [x] Docs only — no workflow, gate, or image change

Refs #989. Does not close it.

### DIFF
--- a/docs/skills/ci/references/failure-modes.md
+++ b/docs/skills/ci/references/failure-modes.md
@@ -28,6 +28,42 @@ The behave summary line (`N scenarios passed, N failed`) and the
 `Failing scenarios:` block name the exact feature file and line. That is the
 only evidence that identifies the failure; job names do not.
 
+## `cap_net_raw` missing on `/usr/bin/ping` is not an image regression
+
+`system_health.feature` "composefs preserves file capabilities on newuidmap,
+newgidmap, and ping" fails deterministically on smoke-b with:
+
+```
+composefs file-capability regression — missing capabilities:
+  ["/usr/bin/ping: expected 'cap_net_raw' in ''"]
+```
+
+This has been triaged twice as a possible bluefin image-content defect. It is
+not one. Fedora's `iputils` deliberately ships `ping` with **no file
+capabilities** — it relies on unprivileged ICMP sockets via
+`net.ipv4.ping_group_range`, which systemd sets by default. Only two binaries
+in that package carry capabilities (`iputils.spec`, `f44`):
+
+```spec
+%attr(0755,root,root) %caps(cap_net_raw=p) %{_bindir}/clockdiff
+%attr(0755,root,root) %caps(cap_net_raw=p) %{_bindir}/arping
+%attr(0755,root,root) %{_bindir}/ping
+```
+
+There is no xattr on `ping` to preserve or strip, so the assertion cannot pass
+on any Fedora-based image regardless of what composefs does.
+
+The same failure proves composefs is working. The other two binaries in the
+assertion do carry capabilities — `shadow-utils` sets `cap_setuid=ep` on
+`newuidmap` and `cap_setgid=ep` on `newgidmap` — and both pass. Capability
+preservation is fine; the expectation list is wrong.
+
+Do not "fix" this by adding `setcap cap_net_raw` to `ping` in the image. That
+diverges from Fedora for no functional gain — `ping` already works. The fix is
+in `projectbluefin/testsuite`: drop `ping` from the assertion, or better, swap
+it for `arping` or `clockdiff`, which genuinely carry `cap_net_raw=p` and so
+make the scenario a real composefs test rather than a vacuous one.
+
 ## The `oras` screenshot error is not the failure
 
 In `projectbluefin/testsuite`'s `e2e.yml@v1`, the `Push desktop screenshot to


Danathar:
Closing — this documents a failure that can no longer happen, and I should have found that before opening it.

`projectbluefin/testsuite#749` ("test(smoke): scope composefs capability check to Fedora-shipped caps") merged **2026-08-09 03:17Z**, about 21 hours before I opened this. It reaches the same conclusion from the same evidence and then actually fixes it:

```diff
-@step("newuidmap, newgidmap, and ping retain their security.capability xattrs")
+@step("newuidmap and newgidmap retain their security.capability xattrs")
     binaries = {
         "/usr/bin/newuidmap": "cap_setuid",
         "/usr/bin/newgidmap": "cap_setgid",
-        "/usr/bin/ping":      "cap_net_raw",
     }
```

with a comment citing the same spec detail — Fedora ships `/usr/bin/ping` plain 0755, only `clockdiff`/`arping` get `%caps(cap_net_raw=p)`, unprivileged ping comes from `net.ipv4.ping_group_range` — and referencing bluefin#989. Their own reference doc records it too.

So the assertion no longer checks `ping`, the deterministic smoke-b failure is gone, and a section in bluefin's triage doc explaining how to diagnose it is now noise for whoever reads it next. The suggestion I made — assert on `arping`/`clockdiff` instead — is also already considered upstream: they kept `newuidmap`/`newgidmap`, which carry real capabilities and preserve the dakota#841 regression signal, which is the better call.

The analysis was right; it just wasn't needed. Checking the downstream repo for an existing fix before writing this up would have caught it. Nothing here to merge.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `01843f3`

PR #1089, opened 2026-08-11
docs(ci): document stable-promotion blocker set for #929

mrbobbytables:

`Execute Release` has failed since July 20 (#929) and prior triage comments on
that issue go stale as the specific failing legs shift run to run. This adds
a dated triage section to `docs/skills/ci/references/failure-modes.md`
recording the current blocker set, re-verified against run
[31355954836](https://github.com/projectbluefin/bluefin/actions/runs/31355954836)
(2026-08-10T04:35Z).

---

## Changes

- The previously-tracked composefs `cap_net_raw` regression (`testsuite#524`)
  **no longer reproduces** in this run — that blocker looks resolved.
- `smoke-a` still fails on the known, unowned Dash-to-Dock/AT-SPI issue.
- `common-b` now fails on a **new** scenario on both `bluefin` and
  `bluefin-nvidia`: `ujust toggle-updates ... non-interactively` hits `gum`'s
  TTY error (`could not open a new TTY`), even though `common#966` (merged
  2026-08-09) should make `ACTION=enable/disable/cancel` skip `gum` entirely.

I reproduced `common@main`'s current `update.just` locally with a real `just`
binary; all three actions (`enable`/`disable`/`cancel`) resolve correctly
without invoking `gum`, so the recipe source on `common@main` alone is not
proven at fault. This sandbox has no podman/skopeo/container tooling to
inspect the actual `common` digest baked into the tested `bluefin:testing`
image, so the section records the open question and the concrete next
diagnostic step (confirm the layered `common` digest) rather than guessing at
a fix. Documentation-only change; no workflow or recipe behavior modified.

Ref: #929

## Testing

- [x] `just check` passes (`Checking syntax: Justfile`)
- [ ] `pre-commit run --all-files` passes (pre-commit not installed in this sandbox; ran the underlying `python3 .github/scripts/validate-docs.py` directly — `documentation ok: 13 skills, 41 Markdown files`)
- [ ] Build tested locally (not applicable — docs only)

### DIFF
--- a/docs/skills/ci/references/failure-modes.md
+++ b/docs/skills/ci/references/failure-modes.md
@@ -45,3 +45,57 @@ That step is `continue-on-error: true`, so it produces a red `##[error]` line in
 the log without failing the job. Runs that pass `:testing` by tag (for example
 `nightly.yml`) do not show it at all. Do not report it as the cause of a
 `post-testing-e2e` failure; find the behave summary instead.
+
+## Stable promotion blocker set (issue #929) — re-verified 2026-08-10
+
+`Execute Release` has failed on every attempt since July 20 (tracked in #929).
+The failing legs move over time; re-check the current run rather than trusting
+an older triage comment. State as of run
+[31355954836](https://github.com/projectbluefin/bluefin/actions/runs/31355954836)
+(2026-08-10T04:35Z, candidate `bluefin:testing` ->
+`sha256:bf615b200faefc44b50232ecc8a3eb21490e88dd9316b528b27f54472122611d`):
+
+| Leg | Status | Notes |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


castrojo (approved):

hanthor (approved):
Docs-only dated triage snapshot for `#929`. It appropriately distinguishes what's confirmed (the `common@main` recipe source itself is not at fault, reproduced locally with a real `just` binary) from what's an open question requiring container tooling this environment doesn't have (which `common` digest is actually layered into the tested image) — and states that distinction honestly rather than guessing at a fix. The composefs `cap_net_raw` blocker is correctly marked resolved-as-of-this-run rather than closed outright. No workflow or recipe behavior changed. (Note: castrojo already approved this on 2026-08-13 for Hive auto-merge; adding an independent review since hanthor hadn't reviewed yet.)

---

PR #1057, opened 2026-08-08
fix(packages): add fzf to base package manifest (#965)

hanthor:
Closes #965

## Summary

Fixes part 1 of #965 where `ujust --choose` fails on fresh installs due to missing `fzf` binary.

- Added `fzf` to `build_files/packages/base.toml` under `[fedora]` section so it is installed in the base image.
- Added `fzf` to `IMPORTANT_PACKAGES` verification list in `build_files/base/20-tests.sh`.

Issue #965, opened 2026-08-03
Missing user settings in Dakota alpha 5


znmeb:
### Before submitting

- [x] I searched existing issues and this is not a duplicate
- [ ] I ran ujust report (or am about to — paste the URL below)

### What happened?

1. `ujust --choose` doesn't work after a fresh install because `fzf` isn't installed.
2. `bluefinctl` is not installed.
3. `ujust dx-group` does not add you to `docker`, `libvirt` or `incus-admin`.

```
Appending docker to /etc/group
grep: /usr/lib/group: No such file or directory
usermod: group 'docker' does not exist
Appending incus-admin to /etc/group
grep: /usr/lib/group: No such file or directory
usermod: group 'incus-admin' does not exist
Appending libvirt to /etc/group
grep: /usr/lib/group: No such file or directory
usermod: group 'libvirt' does not exist
Reboot system and log back in to use docker, libvirt, incus, and serial connections.
```

### Steps to reproduce

1. Install Dakota Alpha 5.
2. Bootc switch to `dakota-gaming:testing`
3. Do `ujust --choose`. You will get an error because `fzf` is not installed.
4. `brew install fzf`, then `ujust --choose` again. Move up to `dx-group`. You will be added to the `dialout` group but not to the others it says you were added to.

Danathar:
## Triage: item 1 is fixed here, items 2 and 3 are Dakota-side

Worked this against `projectbluefin/bluefin@testing`. Splitting the three reports, because only the first one was ever this repo's:

**1. `fzf` missing → `ujust --choose` fails — fixed here.** #1057 added `fzf` to `build_files/packages/base.toml` and to `IMPORTANT_PACKAGES` in `20-tests.sh`. Worth noting it composes correctly with common's `ujust` wrapper rather than duplicating it — the wrapper only brew-bootstraps `fzf` when it is absent:

```bash
if [[ "${CHOOSE}" == true ]] && ! command -v fzf >/dev/null 2>&1; then
    ... brew install fzf ...
```

so the RPM short-circuits the bootstrap, and `--choose` works on first boot before `brew-preinstall.service` has ever run.

[...]

PR #1090, opened 2026-08-11
test(build): regression-guard missing IMPORTANT_PACKAGES entries (#965)

mrbobbytables:

#965 reported (item 1) that `fzf` was missing from Dakota alpha 5, breaking `ujust --choose` on first boot. That was traced back to bluefin and fixed in #1057, which added `fzf` to both `build_files/packages/base.toml` and the `IMPORTANT_PACKAGES` array in `build_files/base/20-tests.sh`.

While verifying that fix, I found there is no regression test asserting the *behavior* the fix depends on: that `20-tests.sh` actually fails the build when `rpm -q` reports a required package absent. The existing `tests/unit/20-tests_test.bats` only has:

- a happy-path test where every package is treated as installed, and
- `"rejects an unwanted package"`, which exercises the separate `UNWANTED_PACKAGES` list, not `IMPORTANT_PACKAGES`.

So nothing in CI currently protects the `IMPORTANT_PACKAGES` gate itself — a future edit could silently drop `fzf` (or any other entry) from the array, or a change to the loop, and no test would catch it, reproducing #965 item 1 with a build that still goes green.

## Changes

Adds one bats test to `tests/unit/20-tests_test.bats`:

- `"20-tests: rejects an image missing a required IMPORTANT_PACKAGES entry"` — stubs `rpm -q` to report `fzf` (and the other packages already treated as absent for `UNWANTED_PACKAGES`) missing, runs the patched script, and asserts a non-zero exit with the `Missing package: fzf` message the script emits.

No production code changed — `build_files/base/20-tests.sh` and `build_files/packages/base.toml` already have the #1057 fix; this only closes the test gap around it.

## Validation

```
$ bats tests/unit/20-tests_test.bats
1..4
ok 1 20-tests: passes when required image state is present
ok 2 20-tests: rejects an image missing a required IMPORTANT_PACKAGES entry
ok 3 20-tests: rejects an unwanted package
ok 4 20-tests: checks NVIDIA packages for NVIDIA images
```

Refs #965

### DIFF
--- a/tests/unit/20-tests_test.bats
+++ b/tests/unit/20-tests_test.bats
@@ -84,6 +84,32 @@ teardown() {
     [ "$status" -eq 0 ]
 }
 
+@test "20-tests: rejects an image missing a required IMPORTANT_PACKAGES entry" {
+    # #965 item 1: fzf was missing from the image, breaking `ujust --choose` on
+    # first boot. #1057 added fzf to IMPORTANT_PACKAGES, but nothing asserted
+    # that the script actually fails when rpm reports it (or any other
+    # required package) absent — the existing "unwanted package" test below
+    # only exercises the UNWANTED_PACKAGES list. Guard the missing-package path
+    # directly so a future edit can't silently drop an entry with no coverage.
+    cat > "${STUB_BIN}/rpm" <<'EOF'
+#!/usr/bin/bash
+if [[ "$*" == *"%{VENDOR}"* ]]; then
+    echo "negativo17.org"
+    exit 0
+fi
+case "$*" in
+    *fedora-flathub-remote*|*fedora-logos*|*fedora-third-party*|*gnome-software*|*podman-docker*) exit 1 ;;
+    *fzf*) exit 1 ;;
+esac
+exit 0
+EOF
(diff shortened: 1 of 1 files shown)


hanthor (approved):
Verified the added bats test against the actual `build_files/base/20-tests.sh` and the existing `20-tests_test.bats` harness. The `IMPORTANT_PACKAGES` loop runs before the `UNWANTED_PACKAGES` loop and iterates in array order (`fzf` comes after `firefox`/`fish`/`flatpak`, all of which the stub treats as present), so the new stub — which fails `rpm -q` only for the existing unwanted-package patterns plus `fzf` — correctly exercises the `Missing package: fzf` exit path added in `#1057`, and does so before the unwanted-package loop could mask it. This closes a real coverage gap: previously nothing asserted the `IMPORTANT_PACKAGES` gate actually fails a build, only the separate `UNWANTED_PACKAGES` path was tested. Test-only change, no production code touched.

---

castrojo (approved):
Reviewed: focused diff matching its stated scope, checks green.

castrojo (approved):
Reviewed: focused diff matching its stated scope, checks green.

PR #1095, opened 2026-08-11
docs(ci): document that :testing can freeze for weeks, masking triage (#929)

mrbobbytables:

`Execute Release` (stable promotion) has failed continuously since July 20 (#929), and prior
triage in that issue kept re-diagnosing the same set of Execute Release symptoms without
asking how stale the image under test actually was.

I verified directly against the GHCR registry (not cached workflow logs) that
`ghcr.io/projectbluefin/bluefin:testing` currently resolves to a digest whose own OCI labels
say `org.opencontainers.image.version: testing-44.20260804` — a build from **2026-08-04**,
a week behind `main`/`testing` HEAD as of today.

Root cause: `build-image-testing.yml` never moves the mutable `:testing` tag itself
(`publish_stream_tag: "false"`); only `promote-to-testing` in `post-testing-e2e.yml` does,
gated on every `run-e2e` leg passing. I paginated `gh api .../actions/workflows/<id>/runs`
for `Post-Testing E2E` from 2026-07-11 through today (~300 runs) and checked each run's
`promote-to-testing` job conclusion directly — it is `skipped` in every single one, because
`smoke-a` (the known, previously-unowned Dash-to-Dock/AT-SPI/Firefox failure discussed in
#929) fails every run. So `:testing` has been unable to advance for roughly a month,
regardless of how many fixes land and build successfully as version-alias tags.

This also explains a symptom flagged in #929/#1089 as an open question: the `toggle-updates`
`ACTION` regression (`gum`'s TTY error) that `common#966` fixed on 2026-08-09 kept
reproducing in Execute Release runs afterward. I pulled the actual `ghcr.io/projectbluefin/common`
layer for the digest built right after that merge and extracted `update.just` — the fix is
correct and skips `gum` entirely for `enable`/`disable`. It just never reached the tested
`bluefin:testing` image, because that image predates the fix by five days.

## Changes

Documentation-only addition to `docs/skills/ci/references/failure-modes.md`: a new section
explaining that `:testing` can silently freeze for weeks whenever any `run-e2e` leg
consistently fails, with copy-pasteable commands to check (a) the actual image digest/build
date Execute Release tested and (b) `promote-to-testing`'s real per-run conclusion, not just
the workflow's overall conclusion — so future triage on stable-promotion failures checks
image freshness before re-attributing the same frozen-image symptom to whichever leg is
easiest to blame that day.

No workflow or recipe behavior is changed by this PR.

Ref: #929

## Testing

- [x] `just check` passes (`Checking syntax: Justfile`)
- [x] `python3 .github/scripts/validate-docs.py` passes (`documentation ok: 13 skills, 41 Markdown files`; `pre-commit` itself is not installed in this sandbox)
- [ ] Build tested locally (not applicable — docs only)

## Checklist

---
🐝 **Hive Agent**: `contributor` | **SHA:** `3487a2b`

### DIFF
--- a/docs/skills/ci/references/failure-modes.md
+++ b/docs/skills/ci/references/failure-modes.md
@@ -112,3 +112,46 @@ Next step: confirm which `common` digest is actually layered into the
 already includes `common#966`, escalate as a `projectbluefin/common` or
 `projectbluefin/testsuite` bug rather than assuming a stale layer will
 self-resolve on the next build.
+## `:testing` can silently freeze for weeks, invalidating every downstream triage (#929)
+
+`build-image-testing.yml` never moves the mutable `:testing` tag itself
+(`publish_stream_tag: "false"`); only `promote-to-testing` in
+`post-testing-e2e.yml` does, and only when every `run-e2e` matrix leg passes.
+If one leg (for example `smoke-a`) fails on every run, `promote-to-testing`
+is `skipped` every time and `:testing` stops advancing — silently, with no
+failed check on the tag itself, because the workflow that would have moved
+it never runs the promotion job at all.
+
+This means `Execute Release`'s `gh api .../manifests/testing` lookup can
+resolve to a build that is *days or weeks* old relative to `main`/`testing`
+HEAD, even though intervening fixes merged and built successfully as
+version-alias tags. Triaging a stable-promotion failure by only reading the
+latest `Execute Release` log reproduces the same symptoms release after
+release and looks like the fixes "didn't work," when the real problem is
+that the tested image predates them. Confirm the actual age of the image
+under test before attributing a release-gate failure to a specific fix:
+
+```bash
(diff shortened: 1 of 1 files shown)


hanthor (approved):
Docs-only addition to `docs/skills/ci/references/failure-modes.md`. Root cause is well-argued: `build-image-testing.yml` never moves the mutable `:testing` tag, only `promote-to-testing` does, and that job is gated on every `run-e2e` leg passing — so a single persistently-failing leg (`smoke-a`) silently freezes `:testing` for weeks with no failing check on the tag itself. The included diagnostic commands (image digest/build-date lookup, `promote-to-testing` per-run conclusion via `gh run view --json jobs`) are copy-pasteable and correct in shape. No workflow or recipe behavior is touched, consistent with the stated scope. This directly explains the `#929`/`#1089` open question about the `toggle-updates`/`gum` fix not reaching the tested image.

---

PR #1100, opened 2026-08-13
fix(lab-check): find the existing check run on busy commits

### LINKED ISSUE #939: ci: grant MergeRaptor app Checks write permission for Lab check

Danathar:
## What

`.github/workflows/lab-check.yml` decides between creating and updating the `testing-lab / <product>` check run by listing check runs on the commit and filtering client-side:

```bash
gh api --method GET "repos/${GITHUB_REPOSITORY}/commits/${SHA}/check-runs" \
  -f per_page=100 \
  --jq '.check_runs | map(select(.name == ... and .app.slug == ...)) | sort_by(.id) | last | .id // empty'
```

That reads **one page of 100 check runs of every name**, and the endpoint returns them **newest first**.

## Why it breaks

Bluefin commits accumulate far more than 100 check runs:

```console
$ for sha in $(git rev-list -n 4 upstream/main); do
    gh api --method GET "repos/projectbluefin/bluefin/commits/$sha/check-runs" \
      -f per_page=1 --jq '.total_count'
  done
227
257
50
57
```

On `2f4dada` the first page of 100 spans `2026-08-13T20:05` back to `2026-08-11T20:17` — about two days — and contains only 22 distinct names; names such as `Scorecard analysis` and `Vulnerability scan / …` exist only on later pages.

The lab check is created early in a commit's life, so once ~100 newer check runs land on that commit it falls off page 1. `CHECK_ID` then comes back empty and the script **POSTs a new check run instead of PATCHing the existing one**. Each lifecycle event (`queued` → `in_progress` → `completed`) creates another duplicate `testing-lab / <product>` entry, and the earlier ones are left stranded in `in_progress` — a check that never completes, which is worse than the missing one.

## The fix

Pass `check_name` so the endpoint filters server-side. Verified against the live API:

```console
$ gh api --method GET "repos/projectbluefin/bluefin/commits/2f4dada.../check-runs" \
    -f check_name="Update MergeRaptor check" -f per_page=100 \
    --jq '{total: .total_count, names: (.check_runs|map(.name)|unique)}'
{"total":39,"names":["Update MergeRaptor check"]}
```

The response then holds only same-named runs, and the existing `sort_by(.id) | last` still picks the newest. `--paginate` is deliberately not added: it would run the `--jq` expression once per page and emit several ids.

## Tests

This path **has never executed in CI**. Every Lab check dispatch since the workflow landed fails at the MergeRaptor token mint (#939) — the last 12+ runs, most recently `31730816503` (2026-08-13T18:27Z) — so the bug could not have been observed from a run log.

`tests/test_lab_check.py` extracts the step's shell body from the workflow and runs it against a stub `gh` that reproduces the endpoint's newest-first ordering, `per_page` truncation and `check_name` filtering:

- update path — the lab check sits behind 150 newer check runs; asserts a single `PATCH` to the existing id, and that `head_sha` is omitted (immutable on an existing run)
- create path — asserts `POST` with `head_sha`, `status`, and no `conclusion` while `in_progress`
- a same-named check run from a different app is ignored

Without the workflow change the first test fails with `AssertionError: 'POST' != 'PATCH'`.

```console
$ python3 -m unittest discover -s tests -p 'test_*.py'
Ran 8 tests in 0.567s

OK
$ actionlint .github/workflows/lab-check.yml   # clean
```

## Relationship to #939

This does **not** resolve #939 — that still needs an org owner to declare and grant **Checks: write** on the MergeRaptor app, which is UI-only and unautomatable. It fixes a defect on the other side of that grant: as written, the first thing to happen after the permission lands would be duplicate lab checks. It is independent of #1079 (which corrects the failure *message*) and touches no line that PR touches.

Refs #939

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/lab-check.yml
+++ b/.github/workflows/lab-check.yml
@@ -82,9 +82,15 @@ jobs:
           STARTED_AT=$(jq -r '.started_at // empty' <<<"${CHECK_JSON}")
           COMPLETED_AT=$(jq -r '.completed_at // empty' <<<"${CHECK_JSON}")
 
+          # check_name filters server-side. Without it the first page is 100
+          # check runs of every name, newest first, so on a busy commit the
+          # existing lab check falls off the page and this lookup returns
+          # nothing -- which silently turns every update into a duplicate
+          # check run (bluefin#939).
           CHECK_ID=$(
             gh api --method GET \
               "repos/${GITHUB_REPOSITORY}/commits/${SHA}/check-runs" \
+              -f check_name="${CHECK_NAME}" \
               -f per_page=100 \
               --jq '.check_runs
                 | map(select(.name == "'"${CHECK_NAME}"'" and .app.slug == "'"${APP_SLUG}"'"))
--- a/tests/test_lab_check.py
+++ b/tests/test_lab_check.py
@@ -1,0 +1,195 @@
+"""Tests for the check-run reporting script in lab-check.yml.
+
+The workflow has never run to completion in CI -- every dispatch so far dies at
+the MergeRaptor token mint (bluefin#939) -- so the script below is exercised
+here against a stub ``gh`` instead.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import subprocess
+import tempfile
+import textwrap
+import unittest
+from pathlib import Path
+
+
+WORKFLOW = Path(__file__).parents[1] / ".github" / "workflows" / "lab-check.yml"
+STEP = "      - name: Create or update lab check\n"
+REPOSITORY = "projectbluefin/bluefin"
+CHECK_NAME = "testing-lab / bluefin"
+SHA = "a" * 40
+
+# Stands in for `gh api`: serves the check-runs listing from GH_STUB_CHECK_RUNS
+# with the same semantics as the real endpoint -- newest first, truncated to
+# per_page, filtered by check_name when asked -- and records every request.
+STUB_GH = '''#!/usr/bin/env python3
+import json
+import os
+import subprocess
+import sys
+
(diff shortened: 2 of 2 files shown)


hanthor (approved):
Solid, well-evidenced fix. The root-cause analysis (check-runs endpoint returns newest-first, unfiltered, and busy Bluefin commits blow past the 100-per-page limit within ~2 days, so the client-side name/app.slug filter never sees the existing lab check and every update POSTs a duplicate instead of PATCHing) is backed by real API output showing 227-257 check runs on recent commits. The fix — passing `check_name` server-side — is minimal and correct, and `sort_by(.id) | last` still correctly picks the newest match afterward. The new `tests/test_lab_check.py` stubs `gh` faithfully (newest-first ordering, per_page truncation, check_name filtering) and covers the update/create/cross-app-name paths; the PR correctly notes this path has never executed in CI due to the separate #939 token issue, which is why the offline test harness is the only verification available. CI is green.

---

PR #1132, opened 2026-08-24
fix(framework): drop user-setup hook that installs a nonexistent brew formula

### LINKED ISSUE #879: Battery Health controls no longer work after upgrading for AMD Framework laptop
### Before submitting

- [x] I searched existing issues and this is not a duplicate
- [x] I ran ujust report (or am about to — paste the URL below)

### ujust report gist URL

https://gist.github.com/mrbobbytables/6439a0a7b3bb09cd02bab1ece9a20305

### What happened?

I have an AMD Framework laptop with a Ryzen 7 7840U w/ Radeon  780M Graphics

On June 9th I updated bluefin, after which I could no longer manage the health of my battery using the [Battery Health Charging Extension](https://maniacx.github.io/Battery-Health-Charging/).  It reported that the [framework kernel module](https://github.com/DHowett/framework-laptop-kmod)  and [framework_tools](https://github.com/FrameworkComputer/framework-system) could not be found. 

The kernel module shouldn't be required anymore with it being upstreamed (I believe).

The tools I was able to install via brew, but still had the issue. I spoke with @castrojo and he found that the Intel framework rules were also being applied to AMD hardware and suggested I do the following:

`sudo rpm-ostree kargs --delete=module_blacklist=hid_sensor_hub && sudo reboot`

Unfortunately that did not fix it, and he suggested I post here directly.

### Steps to reproduce

_No response_

Danathar:
## What

Removes `system_files/shared/usr/share/ublue-os/user-setup.hooks.d/20-framework.sh` and its unit test. The hook's only action is `brew install fw-ectool`, and **`fw-ectool` is not a Homebrew package.**

## Why

Found while re-checking #879 (Framework AMD battery health controls). The reporter said the Battery Health Charging extension could not find `framework_tools`, and that they had to install tooling by hand. bluefin ships a hook whose entire job is to install the Framework EC tool — and it has never worked.

Verified four ways:

| check | result |
|---|---|
| `formulae.brew.sh/api/formula/fw-ectool.json` | `404` |
| `formulae.brew.sh/api/cask/fw-ectool.json` | `404` |
| `Formula/f/fw-ectool.rb` in homebrew/core history | never existed (no commits) |
| `fw-ectool` anywhere in `projectbluefin` org | this line only — no tap provides it |

So `brew install fw-ectool` always errors. The hook runs under `set -euo pipefail`, so it aborts.

The failure is also **invisible after first boot**. `version-script` in `common`'s `libsetup.sh` writes the version *before* the hook body runs:

```bash
# libsetup.sh — records the version, then returns 0 so the body executes
jq ".version.${TYPE_OF_SERVICE}.\"${TARGET_VERSIONING_NAME}\" = \"${VERSION}\"" ... && mv "${tmp}" "${SETUP_CHECKER_FILE}"
```

So `user.20-framework = "1"` is recorded on the first (failing) pass, and every later boot short-circuits at `version-script 20-framework user 1 || exit 0`. Net effect: fails once on Framework hardware, then silently does nothing forever.

The test did not catch this because it stubs `brew` and only asserts that *some* `install` was called — a bogus formula name passes:

```bash
[[ "$1" == "list" ]] && exit 1
exit 0
...
grep -q "install" "${STUB_BIN}/brew.log"
```

## Why removal rather than a corrected package name

The hook's stated purpose was the Framework EC tool "for hardware management (fan curves, battery charge limit, etc.)". The battery-charge-limit half — the subject of #879 — is now handled in-kernel and needs no EC tool: `fw-charge-control.conf` (from #1019) sets `options cros_charge_control probe_with_fwk_charge_control=1`, which is exactly the module param the driver gates on:

```c
/* drivers/power/supply/cros_charge-control.c */
static bool probe_with_fwk_charge_control;
module_param(probe_with_fwk_charge_control, bool, 0644);
...
if (ret > 0 && !probe_with_fwk_charge_control) {
        dev_info(dev, "Framework charge control detected, preventing load\n");
        return -ENODEV;
}
```

With it set, the driver binds and exposes `charge_control_{start,end}_threshold` in sysfs — the interface the extension prefers over any EC binary.

There is also no correct formula to substitute. The official `framework_tool` (`FrameworkComputer/framework-system`) is not in Homebrew; the only related core formula is the third-party `framework-tool-tui`, a different tool and not an obvious stand-in. **Picking a replacement is a maintainer call, so I have not guessed at one** — happy to switch this PR to a corrected package instead of a removal if you'd rather ship a tool.

This removes no working functionality: there is nothing to preserve.

## Scope / what this does not claim

This does **not** close #879. The battery fix itself (#1019) and its regression test (#1060) are already merged, and #879's remaining step is confirming it on a shipped image — still blocked by the `:testing` promotion freeze (#989 → `projectbluefin/testsuite#741`, both still open; `:stable` is `testing-44.20260720`, `:testing` is `testing-44.20260804`, both predating #1019). That is unrelated to this cleanup.

## Testing

- `just check` — clean.
- `bats tests/unit/` via `docker.io/bats/bats` — `fw-charge-control_test.bats` passes (3/3); no test references the removed files. `20-tests: passes when required image state is present` fails **identically on a pristine `upstream/main` worktree** (the minimal bats image lacks `jq`/`rpm`), so it is pre-existing and not from this change.
- `grep -rn 20-framework` — no remaining references in the repo or in `.github/`.

Refs #879


castrojo (approved):
Verified: fw-ectool 404s on formulae.brew.sh (formula and cask), and the hook's only action was installing it. Clean removal including the test that stubbed brew too loosely. Checks green.

PR #1076, opened 2026-08-09
ci: gate stable promotion to Tuesday window

Danathar:
## What does this change?

Gate the testing-to-main promotion's merge-queue enqueue step to the Tuesday
UTC stable release window while keeping the daily promotion heartbeat.

## Why?

Issue #1066 establishes Tuesday as Bluefin's weekly stable release day. The
current caller invokes the shared promotion workflow daily and unconditionally
enqueues the promotion PR, allowing stable releases whenever the merge queue
accepts it. This change adds a caller-side release-window job:

- Tuesday scheduled runs enqueue the promotion PR.
- Manual `workflow_dispatch` remains a documented hotfix escape hatch.
- Push events and non-Tuesday scheduled runs continue refreshing the promotion
  PR and its gates without releasing stable.

The existing signing, E2E, promotion, and merge-queue logic remains in the
shared reusable workflow.

## Validation

- `actionlint .github/workflows/*.yml`
- YAML parse and release-window/`needs` contract assertion
- `just check`
- `python3 .github/scripts/validate-docs.py`
- `git diff --check`
- `pre-commit` was unavailable in the environment (`command not found`)

Refs #1066

— hive: backend=agy agy=1.1.22

Issue #1066, opened 2026-08-09
Weekly stable release: promote testing → stable on Tuesdays


castrojo:
## What we want

Set a **weekly `:stable` release day for Bluefin: Tuesdays**, as part of a coordinated factory-wide 3×-per-week promotion schedule:

| Image | Stable day |
|---|---|
| `bluefin:stable` / `bluefin-nvidia:stable` | **Tuesday** |
| `bluefin-lts:stable` | Thursday (tracked in projectbluefin/bluefin-lts) |
| `dakota:stable` | Sunday (tracked in projectbluefin/dakota) |

## Current state

`promote-testing-to-main.yml` runs daily (`cron: '0 4 * * *'`) plus on every push to `testing`, and the promotion PR auto-merges via merge queue with 0 required approvals. In practice this means `:stable` promotes as soon as a maintainer lets the queue merge — there is no enforced weekly cadence.

## Proposed change

- Gate the promotion merge on a Tuesday release window (UTC), e.g. only run the merge-queue enqueue step, or a pre-merge release-window check, when `date +%u == 2`.
- Keep the daily heartbeat so the promotion PR stays fresh and gate checks (`release/ready`, cosign verify) re-run — but only allow the actual merge/release on Tuesdays.
- Keep `workflow_dispatch` as the documented hotfix escape hatch.
- Update `release-reminder.yml` (`days_warn`/`days_escalate`) to match a 7-day cadence if needed.

## Coordination

- Schedule rationale and cross-repo plan: see the common issue (projectbluefin/common, "Weekly stable release cadence").
- Docs to update once this lands: `docs/skills/release-promotion/references/pr-based-release.md` and `docs/skills/image-registry.md` in projectbluefin/common.

castrojo:
Coordinating issue: projectbluefin/common#975 (recommended Tue/Thu/Sun schedule and docs updates).

castrojo:
Resolved by merged PR #1076 (ci: gate stable promotion to Tuesday window) and PR #1134 (test(ci): pin the Tuesday stable release window).

PR #1134, opened 2026-08-24
test(ci): pin the Tuesday stable release window

Danathar:
## What

Adds `tests/test_promote_release_window.py` covering the Tuesday release gate, and documents the non-obvious mechanism the gate depends on. **No behaviour change.**

## Why

The core ask of #1066 landed in #1076 — `:stable` promotes on Tuesdays, the daily run still refreshes the PR, `workflow_dispatch` stays as the hotfix path. I verified all three against the merged workflow. But it shipped with **no test**, and the release day is a bare string comparison inside a shell step:

```bash
weekday=$(date -u +%u)
if [[ "$weekday" == "2" ]]; then
```

reaching the promotion through one expression:

```yaml
use_merge_queue: ${{ needs.release_window.outputs.should_enqueue == 'true' }}
```

A typo in either place moves the release day or disconnects the gate, with nothing failing.

The test extracts the step from the workflow and **executes it** against a stubbed `date`, rather than asserting on its text — so it covers the real shell, including the `case` fallback.

Two assertions cover traps that are invisible by reading:

- **`date -u` is load-bearing.** Without it the release day follows the runner's clock, not the UTC window the cron is written in.
- **`%u` is load-bearing.** It is what makes the literal `"2"` mean Tuesday. Per projectbluefin/common#975 the sibling repos gate on **Thursday** and **Sunday** with this same pattern, and there `%u` (Mon=1..Sun=7) and `%w` (Sun=0..Sat=6) disagree — Sunday is `7` under `%u` and `0` under `%w`. Pinning the convention here is worth more than for Tuesday alone, since this step is the template the other two repos copy.

### Mutation-tested, so the assertions aren't decorative

| Mutation | Result |
|---|---|
| Release day Tuesday → Wednesday | 2 failures |
| Drop `-u` (runner-local clock) | 1 failure |
| `%u` → `%w` | 1 failure |
| `*)` fallback opens the gate | 3 failures |
| `use_merge_queue` pinned to constant `true` | 1 failure |

## The comment, and a follow-up that isn't mine to make

While tracing the gate I read `reusable-promote-squash.yml` in `projectbluefin/actions` and found the mechanism is not what the input name suggests:

```bash
if [ "${{ inputs.use_merge_queue }}" = "true" ]; then
  # enqueuePullRequest GraphQL
else
  gh pr merge "$PR_NUMBER" --repo ... --auto 2>/dev/null \
    && echo "✅ Auto-merge enabled" \
    || echo "⚠️  Auto-merge not available — maintainer merges when checks pass"
fi
```

So `use_merge_queue: false` does **not** mean "do not merge" — it routes to `gh pr merge --auto`, which the merge queue ruleset on `main` rejects. Nothing merges, which is the intended outcome, but it arrives via a path that is blocked rather than one that declines to act.

That matters because the input documents itself as a statement of fact — *"Set true when the target branch is protected by a GitHub merge queue ruleset"* — and `main` always **is** so protected. A reader tidying this up to a constant `true` would be acting on a true statement and would silently delete the release window. The existing comment explains why `true` is used but says nothing about `false` carrying the gate. This PR adds that, and the wiring test backs it up.

**Follow-up for maintainers, not done here:** expressing this directly needs a merge on/off input (e.g. `enable_merge`) in `projectbluefin/actions` — a different repo, and a design call. I have not tried to work around it from this side; a label-juggling or `--disable-auto` shim would add races and complexity for no behavioural gain. Flagging it rather than guessing.

## On the other issue bullets

- **`release-reminder.yml` (`days_warn`/`days_escalate`)** — the issue says "if needed". I checked `reusable-release-reminder.yml`: `days_warn` means *"post a reminder after this many days without a release"*, and the current `7`/`14` already fit a 7-day cadence (warn when a release is due, escalate after a full missed window). Retuning it is a subjective threshold call, so I left it alone rather than change it unilaterally.
- **Docs** (`pr-based-release.md`, `image-registry.md`) are tracked in `projectbluefin/common` per #975, not this repo.

## Unrelated, but worth flagging

Every recent `Promote testing to main` run is failing, before the gate is ever reached:

```
! [remote rejected] auto/promote-testing-to-main -> auto/promote-testing-to-main (protected branch hook declined)
```

The workflow can't push its own promotion branch — a ruleset/permission issue on `auto/promote-testing-to-main`, not something a PR can fix, and unrelated to this change. Mentioning it because the Tuesday gate is academic until it's resolved.

## Testing

- `python3 -m unittest discover -s tests -p 'test_*.py'` — 16 tests, all pass.
- Mutation results above.
- `actionlint` on the workflow — clean, exit 0.
- `just check` — clean.

Refs #1066

### DIFF
--- a/.github/workflows/promote-testing-to-main.yml
+++ b/.github/workflows/promote-testing-to-main.yml
@@ -67,5 +67,17 @@ jobs:
       run_e2e: false
       # main uses a merge queue ruleset (17070404): gh pr merge --auto is blocked.
       # use_merge_queue: true calls enqueuePullRequest GraphQL instead.
+      #
+      # This input doubles as the release gate, so read false carefully: it does
+      # not mean "do not merge", it routes the reusable workflow to
+      # `gh pr merge --auto`, which the merge queue ruleset above rejects -- so
+      # nothing merges. That is what holds :stable to Tuesdays (bluefin#1066)
+      # while the daily run still refreshes the PR and re-runs the gate checks.
+      #
+      # Do not "correct" this to a constant true on the grounds that main is
+      # always queue-protected: that statement is true and it would delete the
+      # release window. Expressing this directly needs a merge on/off input in
+      # projectbluefin/actions; until that exists,
+      # tests/test_promote_release_window.py pins the wiring.
       use_merge_queue: ${{ needs.release_window.outputs.should_enqueue == 'true' }}
     secrets: inherit
--- a/tests/test_promote_release_window.py
+++ b/tests/test_promote_release_window.py
@@ -1,0 +1,171 @@
+"""Tests for the stable release window gate in promote-testing-to-main.yml.
+
+`:stable` is meant to promote on Tuesdays only (bluefin#1066). The gate that
+enforces that is a shell step whose output decides whether the promotion PR is
+enqueued for merge, and it shipped with no coverage -- so nothing catches a
+typo that moves the release day or opens the gate on every run.
+
+The step is extracted from the workflow and executed here against a stubbed
+`date`, so these tests exercise the real shell rather than asserting on its
+text.
+"""
+
+from __future__ import annotations
+
+import os
+import subprocess
+import tempfile
+import textwrap
+import unittest
+from pathlib import Path
+
+
+WORKFLOW = (
+    Path(__file__).parents[1] / ".github" / "workflows" / "promote-testing-to-main.yml"
+)
+STEP = "      - name: Determine whether to enqueue the promotion\n"
+
+# Stands in for `date`: records the arguments it was called with, and answers a
+# weekday request from GH_STUB_WEEKDAY. Anything else falls through to the real
+# binary so the step is not silently reshaped by the stub.
+STUB_DATE = '''#!/usr/bin/env python3
+import os
+import subprocess
+import sys
+
+arguments = sys.argv[1:]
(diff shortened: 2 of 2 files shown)


hanthor (approved):
Confirmed this is test+docs only — the diff to `promote-testing-to-main.yml` is comments alone; the actual `use_merge_queue: ${{ needs.release_window.outputs.should_enqueue == 'true' }}` line is unchanged.

The test approach is the right one for a bash-in-YAML gate: it extracts the real step body and executes it against a stubbed `date` rather than pattern-matching the YAML text, so it actually catches a broken `-u`/`%u`/fallback rather than just a moved string. The `%u` vs `%w` note is a genuinely useful catch (Sunday is `7` vs `0` between them) given this step is described as the template the other two repos copy. Mutation table covers the traps that matter: wrong day, dropped `-u`, wrong format specifier, permissive fallback, and the `should_enqueue`↔`use_merge_queue` wiring itself.

Also worth the maintainers' attention: the PR flags (out of scope, correctly not touched here) that every recent `Promote testing to main` run is failing before the gate even runs — `protected branch hook declined` pushing `auto/promote-testing-to-main`. Worth a follow-up since it makes this gate moot until resolved.

CI green (Unit tests, validate, Check PR base branch).

---

castrojo (approved):

PR #1136, opened 2026-08-24
docs(ci): record Testing Images build failing since 2026-08-12 (#995)

Danathar:
## Summary

Re-verifies issue #995's audit as of 2026-08-24. The previously-tracked
blocker (firefox.feature/AT-SPI e2e failure, #989, last re-verified
2026-08-10) is no longer the proximate cause of `:testing` staying frozen.

**New finding:** the `Testing Images` build itself has failed 9/9 times since
2026-08-12 — 12 straight days, zero successes — with an rpmdb `database disk
image is malformed` error in the `extension-builder` Containerfile stage
(`dnf5 -y install glib2-devel meson sassc cmake dbus-devel`), on both `main`
and `nvidia` flavors, on every retry. Confirmed live against run
[32753582407](https://github.com/projectbluefin/bluefin/actions/runs/32753582407)
(2026-08-24), which I watched to completion for this PR.

Because `post-testing-e2e.yml`'s `e2e` job requires the triggering
`Testing Images` run to have `conclusion == 'success'`, every
`Post-Testing E2E` run since 08-12 now reports `e2e`/`run-e2e`/
`promote-to-testing` as **skipped**, not failed. The AT-SPI bug documented in
the adjacent section is still real and unfixed, but it's no longer what's
stopping `:testing` from advancing — the pipeline doesn't get far enough to
reach it anymore.

`Containerfile` and `image-versions.yml` are unchanged since 2026-08-07
(before the last known-good build on 08-10), ruling out a script/pin
regression as the direct trigger.

I flag an **unconfirmed** prime suspect for whoever picks this up next: the
registry-based buildah layer cache (`Justfile`'s `cache_ref`) is keyed only
by image flavor, not by tag, and is written concurrently by `testing`- and
`main`-branch builds of the same flavor (`promote-testing-to-main` triggers a
`main` build minutes after a `testing` push). I could not confirm or rule
this out — no CI trigger access from this runtime, and per this repo's
`AGENTS.md` a documentation-only change should not run an expensive image
build. Left as a concrete next step for a maintainer with a
cache-disabled test build.

This is a documentation-only update to `docs/skills/ci/references/failure-modes.md`,
matching the pattern of prior audit follow-ups on #995 (#1013, #1049, #1092).
No GHCR mutation, no workflow change, no attempt to fix or bypass a gate.

## Test plan

- [x] `python3 .github/scripts/validate-docs.py` passes
- [x] No trailing whitespace / file ends with newline
[...]

Issue #995, opened 2026-08-07
audit: :stable and :testing point at images that never passed e2e


castrojo:
## Summary

**Users tracking `ghcr.io/projectbluefin/bluefin:stable` are running an image that never passed e2e.** So are users on `:testing`. This is a follow-up audit to the `publish_stream_tag` stream-tag leak fixed in projectbluefin/actions#405.

This issue is an **audit report** — nothing in GHCR has been re-tagged or deleted, and no release behavior has been changed. It documents what happened so a human can decide the remediation.

### 1. `post-testing-e2e` has not passed since 2026-06-22

`gh run list --repo projectbluefin/bluefin --workflow post-testing-e2e.yml -L 400`:

| Period | Result |
|---|---|
| Last `success` | **2026-06-22** (runs `27922429853`, `27922740150`) |
| 2026-06-23 → 2026-08-07 | **0 success**, ~200 runs, all `failure` or `skipped` |

The `promote-to-testing` job — the only legitimate writer of the `:testing` tag — has therefore not run since 2026-06-22.

### 2. …yet `:testing` kept moving anyway (the leak, now fixed)

`reusable-build.yml`'s `compute-push-tags` published the bare stream tag on every build despite `publish_stream_tag: "false"`, because `just generate-build-tags` emits `testing` as `alias_tags[0]`. Fixed in projectbluefin/actions#405 (`43c34f6`), live on `@v1`.

Registry proof that `:testing` was written by the **build**, not by a promotion — all three tags sit on one package version pushed at build time:

```
$ gh api "/orgs/projectbluefin/packages/container/bluefin/versions"
sha256:bf615b200faef  2026-08-04T16:42:15Z  testing-20260804,testing-44.20260804,testing
```

`sha256:bf615b20…` is the digest that **failed** e2e in run [30931237054](https://github.com/projectbluefin/bluefin/actions/runs/30931237054) (`smoke-a` = failure, `promote-to-testing` = **skipped**).

### 3. `:stable` was promoted from that ungated `:testing` pointer, with no gate of its own

`execute-release.yml` / `promote-bluefin-stable.yml` both pass:

```yaml
variants: '[{"image":"bluefin","source_tag":"testing","target_tag":"stable"}, ...]'
```

`reusable-execute-release.yml`'s `resolve` job does `skopeo inspect ghcr.io/projectbluefin/bluefin:testing` — **it trusts whatever `:testing` points at, by design.** `:testing` is the release input.

[...]

PR #1153, opened 2026-08-28
fix(build): checkpoint the rpmdb out of WAL mode at every stage commit

Danathar:
## Summary

Every Testing Images build since 2026-08-13 has died at the same step: the first stage that tries to read the RPM package database written by the stage before it gets back "database disk image is malformed". No image means e2e never runs, which means `:testing` stays frozen — the promotion chain that issue #995 tracks. The cause class is that the RPM database is a SQLite file in write-ahead-log (WAL) mode: after each build stage it is left with companion `-shm`/`-wal` files that the next stage's reader has to reassemble through the container layer filesystem, and in CI that reassembly breaks. This PR makes each stage leave the database as one ordinary, self-contained file (checkpoint the WAL, switch to the default journal mode, delete the sidecars), so the next stage — and the shipped image — reads a plain file with no WAL machinery involved.

## Forensics (what was eliminated)

The earlier audit note (#1136) suspected the flavor-keyed registry layer cache replaying a corrupt layer, and suggested a maintainer trigger a cache-disabled build. Full-log comparison of the last-good and first-bad runs disproves that and everything else that is pinned:

| Input | Last good build ([31607601482](https://github.com/projectbluefin/bluefin/actions/runs/31607601482), 08-12) | First malformed ([31727537250](https://github.com/projectbluefin/bluefin/actions/runs/31727537250), 08-13) |
|---|---|---|
| Runner image | ubuntu-24.04 `20260720.247.2` | identical |
| podman / buildah / crun (resolute) | 5.7.0 / 1.42.1 / 1.21 | identical |
| `projectbluefin/actions` | `370431f`→`8906e13` delta touches only sync-branches/renovate workflows | same `8906e13` |
| bluefin delta between the runs | — | only `20-tests.sh` (+9 lines, runs in Stage 2, after the failing step) |
| Base compose packages | `44.20260812.0` | `44.20260813.0` — **byte-identical `rpm -qa` set**, byte-identical shipped `-shm`/`-wal` sidecars |

Notes that correct the earlier census:

- The 08-12 run **built every stage successfully** (including extension-builder) and failed only on GHCR pushes — secondary rate limit 403s. The malformed-rpmdb signature starts 08-13.
- The registry-cache-replay theory cannot hold: the base digest is re-resolved from `silverblue:44` daily, so Stage 1's cache key changes daily and it rebuilds fresh in every failing run — seven distinct daily base digests (08-13 → 08-26) all fail identically, and run [32924887367](https://github.com/projectbluefin/bluefin/actions/runs/32924887367) (08-26) demonstrably executed Stage 1 live (new `BUILD_FILES_SHA`, full 251-package install in the log) before the child stage's read failed.

What remains after elimination: the Fedora bootc bases ship `/usr/lib/sysimage/rpm/rpmdb.sqlite` in WAL mode with stale sidecars, every dnf transaction re-enables WAL, and the failing read (`SELECT hnum, blob FROM 'Packages'` is the rpmdb Packages table) is the next stage's first open of that committed WAL state under CI's rootful buildah 1.42.1 overlay. The same image and same read succeed under rootless podman 5.8.4 locally, so the residual trigger sits in the buildah/overlay interaction with the post-0812 database bytes — not in any input this repo pins, which is why removing the WAL state at the commit boundary is the right repo-side fix regardless of which side of that interaction is ultimately at fault. (ublue-os/main builds the same bases green today; its Containerfile has no dnf-stage → dnf-stage boundary and it uses the runner's stock podman.)

## Change

- `build_files/shared/checkpoint-rpmdb.sh` — `PRAGMA wal_checkpoint(TRUNCATE)` + `PRAGMA journal_mode=DELETE`, then remove `-shm`/`-wal`; fails loudly if the rpmdb is missing (`sqlite3.connect` would otherwise silently create an empty one).
- `Containerfile` — run it as the last step of Stage 1 (`base-common`) and Stage 2, and add it to the `ctx` COPY list. Per-stage is required, not optional: a stage's own dnf write always flips the database back to WAL (observed directly).
- `tests/unit/checkpoint-rpmdb_test.bats` — real-SQLite tests: converts to rollback-journal mode, removes sidecars, preserves rows, fails without creating a stray db.
- `tests/unit/containerfile_contexts_test.bats` — expected `ctx` input set gains the new script.
- `docs/skills/ci/references/failure-modes.md` — records the corrected census and retires the cache-disabled-run experiment so nobody re-runs a settled question.

## Validation

- Two-stage build on the failing-era base (`44.20260824`, digest `c43ca97f`) with the checkpoint in place: the committed layer carries a single `rpmdb.sqlite`, the next stage reads `journal_mode=delete` with no sidecars present, and `rpm -qa` + a dnf install succeed. The corrupting condition itself only manifests under CI's builder, so the merge-queue / push build of this PR is the end-to-end test — if the fix is wrong, Testing Images stays red and nothing regresses.
- `bats tests/unit/`: 168 tests, the only failures are the two pre-existing non-hermetic framework-hook tests that #1144 fixes (they fail on any host with a real `rpm-ostree`).
- `shellcheck` clean; `python3 .github/scripts/validate-docs.py` passes.

Refs #995 (unfreezes the promotion chain; the audit's re-pointing decisions remain human-gated). Once a Testing Images run goes green, post-testing-e2e will run again and #989's AT-SPI failure becomes the next blocker in line.

— hive: backend=claude model=claude-fable-5

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -35,6 +35,7 @@ COPY /build_files/shared/utils/ghcurl /build_files/shared/utils/ghcurl
 FROM scratch AS ctx
 COPY /system_files /system_files
 COPY /build_files/shared/build-gnome-extensions.sh /build_files/shared/build-gnome-extensions.sh
+COPY /build_files/shared/checkpoint-rpmdb.sh /build_files/shared/checkpoint-rpmdb.sh
 COPY /build_files/shared/clean-stage.sh /build_files/shared/clean-stage.sh
 COPY /build_files/shared/disable-repos.sh /build_files/shared/disable-repos.sh
 COPY /build_files/shared/finalize-gnome-extensions.sh /build_files/shared/finalize-gnome-extensions.sh
@@ -86,7 +87,8 @@ RUN --mount=type=cache,dst=/var/cache/libdnf5 \
         export PATH="/tmp/scripts/helpers:$PATH" && \
         /ctx/build_files/base/03-packages.sh && \
         /ctx/build_files/base/04-install-kernel-akmods.sh && \
-        /ctx/build_files/base/05-override-install.sh \
+        /ctx/build_files/base/05-override-install.sh && \
+        /ctx/build_files/shared/checkpoint-rpmdb.sh \
     '
 
 # hadolint ignore=DL3006
@@ -150,7 +152,8 @@ RUN --mount=type=cache,dst=/var/cache/libdnf5 \
         /ctx/build_files/base/19-initramfs.sh && \
         /ctx/build_files/shared/validate-repos.sh && \
         /ctx/build_files/shared/clean-stage.sh && \
-        /ctx/build_files/base/20-tests.sh \
+        /ctx/build_files/base/20-tests.sh && \
+        /ctx/build_files/shared/checkpoint-rpmdb.sh \
     '
 
 # Embed the Stable container-native ISO contract after Stage 2. This runs
--- a/build_files/shared/checkpoint-rpmdb.sh
+++ b/build_files/shared/checkpoint-rpmdb.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/bash
+
+echo "::group:: ===$(basename "$0")==="
+
+set -eoux pipefail
+
+# Leave the rpmdb as a single self-contained file in the committed layer.
+#
+# The Fedora bootc base images ship /usr/lib/sysimage/rpm/rpmdb.sqlite in
+# SQLite WAL journal mode together with stale rpmdb.sqlite-{shm,wal}
+# sidecars, and every dnf transaction in a build stage switches the database
+# back to WAL and leaves fresh sidecars behind. A layer committed in that
+# state is not self-contained: the next stage's first rpmdb read must
+# reconstruct WAL state through overlayfs, which is where CI's
+# "database disk image is malformed" failures come from (issue #995;
+# docs/skills/ci/references/failure-modes.md). Checkpointing the WAL and
+# switching to the default rollback-journal mode makes the database an
+# ordinary single file that any later stage — and the shipped image — can
+# read without WAL machinery. Run this as the last step of any RUN whose
+# rpmdb a later stage or the final image will read.
+
+# RPMDB_PATH: absolute path of the rpmdb SQLite database.
+# Defaults to the bootc sysimage location; overridden in unit tests.
+RPMDB_PATH="${RPMDB_PATH:-/usr/lib/sysimage/rpm/rpmdb.sqlite}"
+
+# sqlite3.connect() would silently create an empty database at a wrong path;
+# a build without an rpmdb here is broken and must fail now, not later.
+if [[ ! -f "${RPMDB_PATH}" ]]; then
+    echo "checkpoint-rpmdb: no rpmdb at ${RPMDB_PATH}" >&2
+    exit 1
+fi
+
+python3 - "${RPMDB_PATH}" <<'PYEOF'
+import sqlite3
+import sys
+
+db = sqlite3.connect(sys.argv[1])
+db.execute("PRAGMA wal_checkpoint(TRUNCATE)")
+db.execute("PRAGMA journal_mode=DELETE")
+db.close()
+PYEOF
+
+rm -f "${RPMDB_PATH}-shm" "${RPMDB_PATH}-wal"
+
+echo "::endgroup::"
--- a/docs/skills/ci/references/failure-modes.md
+++ b/docs/skills/ci/references/failure-modes.md
@@ -347,3 +347,53 @@ gh run list --repo projectbluefin/bluefin --workflow "Testing Images" -L 20 \
 gh run view RUN_ID --repo projectbluefin/bluefin --json jobs \
   --jq '.jobs[] | [.name, .status, .conclusion] | @tsv'
 ```
+
+### Update 2026-08-28 — registry-cache suspect disproven; failure is a WAL-mode rpmdb read across the stage-commit boundary
+
+Full-log forensics across the good/bad boundary correct two claims above and
+retire the prime suspect:
+
+- **The census conflates two failure modes.** Run `31607601482` (08-12) built
+  every stage successfully — including the `extension-builder` dnf install —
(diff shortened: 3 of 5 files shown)


hanthor (approved):
This is the actual root-cause fix for the rpmdb corruption #1136 documented (and which I'd noted has the same error signature as `bluefin-lts#538`, fixed earlier this week). The forensics table is genuinely rigorous — it systematically eliminates runner image, tool versions, actions-repo delta, and package-set as variables (byte-identical `rpm -qa` and sidecar files between the last-good and first-bad runs), and directly falsifies the earlier registry-cache-replay theory from #1136 with a concrete counter-run (`32924887367`) that rebuilt Stage 1 live and still failed. That's a genuine correction of a prior hypothesis, not just a restatement.

The fix itself is correct: `checkpoint-rpmdb.sh` checks the db file exists *before* calling `sqlite3.connect()` (which would otherwise silently create an empty database at the wrong path and mask a real failure), then `PRAGMA wal_checkpoint(TRUNCATE)` + `journal_mode=DELETE` to collapse WAL state into the single file, then removes the now-stale `-shm`/`-wal` sidecars. Applying it at the end of *both* Stage 1 and Stage 2 (not just once) is correct per the PR's own observation that any stage's dnf write re-flips the db back into WAL — a single end-of-build checkpoint wouldn't catch a Stage 1→Stage 2 read failure. Tests use a real WAL-mode SQLite fixture and verify journal-mode conversion, sidecar removal, content preservation, and the fail-loud-on-missing-db path. `Containerfile`'s `ctx` COPY list and the corresponding bats assertion are both updated together.

One thing worth a maintainer's attention, not a blocker on this PR: `bluefin-lts#538` fixed the same error signature but only checkpoints once, at final `cleanup.sh`, in a single-stage script sequence rather than a multi-stage `Containerfile`. Given this PR's finding that *every* dnf transaction re-enables WAL, it's worth double-checking whether bluefin-lts's build has any intermediate stage boundary that could see the same read failure before final cleanup runs — I can't tell from this PR alone whether that architecture difference makes bluefin-lts's single checkpoint sufficient or not.

CI green (`validate`, `Unit tests`, `Check PR base branch`, `validate-release-notes`); the real end-to-end proof is the next live build, as the PR itself says.

---

KiKaraage:
Currently failing when installing `uupd`? Happen during "Build Testing Images" stage, but also validated by #1166 

https://github.com/projectbluefin/bluefin/actions/runs/33301558385/job/99230611249

```
  + copr_install_isolated ublue-os/packages uupd
  + local copr_name=ublue-os/packages
  + shift
  + packages=('uupd')
  + local packages
  + [[ 1 -eq 0 ]]
  + repo_id=copr:copr.fedorainfracloud.org:ublue-os:packages
  + echo 'Installing uupd from COPR ublue-os/packages (isolated)'
  + dnf5 -y copr enable ublue-os/packages
  Installing uupd from COPR ublue-os/packages (isolated)
   https://copr.fedorainfracloud.org/api_ ???% |   0.0   B/s |   0.0   B |  02m23s
  >>> Curl error (28): Timeout was reached for https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/ [Operation too slow. Less than 1000 bytes/sec transferred the last 30 seconds] - https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/
  >>> Curl error (28): Timeout was reached for https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/ [Operation too slow. Less than 1000 bytes/sec transferred the last 30 seconds] - https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/
  >>> Curl error (56): Failure when receiving data from the peer for https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/ [OpenSSL SSL_read: OpenSSL/3.5.7: error:0A000126:SSL routines::unexpected eof while reading, errno 0] - https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/
  >>> Curl error (28): Timeout was reached for https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/ [OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 0] - https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/
  >>> Curl error (28): Timeout was reached for https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/ [OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 0]
  Failed to download files
   Librepo error: Curl error (28): Timeout was reached for https://copr.fedorainfracloud.org/api_3/rpmrepo/ublue-os/packages/fedora-44/ [OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 0]
  Error: building at STEP "RUN --mount=type=cache,dst=/var/cache/libdnf5 --mount=type=cache,dst=/var/cache/rpm-ostree --mount=type=bind,from=ctx-build,source=/build_files,target=/ctx/build_files --mount=type=bind,from=ctx-build,source=/image-versions.yml,target=/ctx/image-versions.yml --mount=type=secret,id=GITHUB_TOKEN --mount=type=tmpfs,dst=/boot bash -euo pipefail -c '         echo "build_files cache key: ${BUILD_FILES_SHA}" &&         dnf5 config-manager setopt keepcache=1 &&         dnf5 config-manager setopt install_weak_deps=0 &&         dnf5 -y swap fedora-logos generic-logos &&         rpm --erase --nodeps --nodb generic-logos &&         mkdir -p /tmp/scripts/helpers &&         install -Dm0755 /ctx/build_files/shared/utils/ghcurl /tmp/scripts/helpers/ghcurl &&         export PATH="/tmp/scripts/helpers:$PATH" &&         /ctx/build_files/base/03-packages.sh &&         /ctx/build_files/base/04-install-kernel-akmods.sh &&         /ctx/build_files/base/05-override-install.sh &&         /ctx/build_files/shar
  error: recipe `build` failed with exit code 1
  error: recipe `build-ghcr` failed with exit code 1
```

PR #1154, opened 2026-08-28
docs(ci): resolve the common-b toggle-updates blocker as a freeze artifact

Danathar:
## Summary

The CI triage runbook has carried an open question since 2026-08-10: `common-b` fails on `ujust toggle-updates` with a `gum` TTY error instead of skipping, with two competing explanations and an explicit instruction not to assume a stale layer would self-resolve. This settles it with evidence. The answer removes one of the two failing legs from the list of things that must be fixed before `:testing` can advance again — it is an artifact of the freeze, not a defect, and nothing in this repo or `common` needs to change for it.

Docs only. No workflow, gate, or image change.

## What the evidence shows

**The image under test predates the contract the scenario asserts.** `ghcr.io/projectbluefin/bluefin:testing` still resolves to the 2026-08-04 build — unchanged since the freeze was first reported on 2026-08-07:

```console
$ skopeo inspect docker://ghcr.io/projectbluefin/bluefin:testing \
  | python3 -c 'import json,sys; d=json.load(sys.stdin); print(d["Labels"]["org.opencontainers.image.version"], d["Created"])'
testing-44.20260804 2026-08-04T16:32:43.940827075Z
```

`common#966` (`93068cd`), which added the non-interactive `toggle-updates ACTION=` contract, merged **2026-08-09T02:16Z** — five days after that build. So the tested image ships the previous `update.just` (`bd90527`, 2026-06-20). Hypothesis 1 confirmed; the SSH/session-harness hypothesis is refuted, since nothing in the harness is involved.

**Why the `@requires_toggle_action` gate never skipped.** The pre-`#966` recipe already declared `toggle-updates ACTION="prompt"` — it accepts the argument and then ignores it, falling straight through to the interactive body. The probe `ujust toggle-updates cancel` therefore reaches `gum choose`, which fails with no TTY and leaves `SELECTED_OPTION` empty, and the recipe's own next line is:

```bash
[[ "${SELECTED_OPTION}" == "Cancel" || "${SELECTED_OPTION}" == "" ]] && exit 0
```

which exits `0`. The gate that exists to skip images lacking ACTION support passes on exactly those images. An exit-status probe cannot detect this contract on any image — it has to observe that `enable` actually changes timer state, or inspect the recipe body. **That is the one real bug left in this thread, and it belongs to `projectbluefin/testsuite`**, not here.

**Why it needs no fix on our side.** This repo's `common` pin in `image-versions.yml` is bumped by Renovate near-daily; it has carried `#966` since #1069 (2026-08-09) and is current as of #1146 (today). Any freshly built image satisfies the scenario, so the leg recovers on its own once promotion resumes.

## Changes

- Rewrites the open-question section as a resolved finding, with the registry evidence, the recipe-level explanation for the non-skipping gate, and an explicit "do not count this as a blocker" note.
- Re-labels both `common-b` rows in the blocker table as freeze artifacts rather than `failing (new)`.
- Extends the existing digest-lookup recipe with the **build date**. It resolved the digest but not its age, and age is the field that settles "is this image older than the fix I am looking for?" — the question this whole section turned on, and the reason it stayed open.

## Scope notes

- Deliberately avoids the region #1153 edits in the same file, so the two do not conflict.
- Does not touch the `smoke-a` / `firefox.feature` blocker, which I re-verified as unchanged in today's nightly ([33142211266](https://github.com/projectbluefin/bluefin/actions/runs/33142211266)): all six scenarios still fail through both retries. That remains testsuite-owned and is correctly documented already.
- Per #989's "Explicitly NOT the fix", nothing here adds `always()`, `continue-on-error`, or alters `promote-to-testing`'s `needs:`.

## Validation

- `python3 .github/scripts/validate-docs.py` — `documentation ok: 13 skills, 41 Markdown files`
- `just check`
- `skopeo inspect` command in the new snippet run as written (output above)
- Registry findings cross-checked a second way, against the raw GHCR manifest/config API, and agree

Refs #989

— hive: backend=claude model=claude-fable-5

### DIFF
--- a/docs/skills/ci/references/failure-modes.md
+++ b/docs/skills/ci/references/failure-modes.md
@@ -71,47 +71,60 @@ an older triage comment. State as of run
 | Leg | Status | Notes |
 |---|---|---|
 | `bluefin` / `smoke-a` | failing | Dash to Dock / Firefox / AT-SPI session lookups fail (`gdbus ... Extensions.GetExtensionInfo` returns `UnknownMethod`, Firefox/Settings not found via AT-SPI). No open PR claims this; see prior findings on #929. |
-| `bluefin` / `common-b` | failing (new) | `ujust toggle-updates` non-interactive scenario — see below. |
-| `bluefin-nvidia` / `common-b` | failing (new) | Same `ujust toggle-updates` failure, identical error, on the NVIDIA variant. |
+| `bluefin` / `common-b` | failing — freeze artifact, not a defect | `ujust toggle-updates` non-interactive scenario. The tested image predates the contract the scenario asserts; resolved below, nothing to fix here. |
+| `bluefin-nvidia` / `common-b` | failing — freeze artifact, not a defect | Same cause on the NVIDIA variant, which is why both fail identically. |
 | `bluefin` / `bluefin-nvidia` smoke-b, common-a | passing | The composefs `cap_net_raw` regression tracked in `testsuite#524` / `dakota#841` **no longer appears** in this run — treat that blocker as resolved unless a fresh run shows it again. |
 
-### New: `ujust toggle-updates` fails with a `gum` TTY error, not a skip
+### Resolved: `ujust toggle-updates` fails because the tested image predates the contract
 
 `tests/common/features/common_ujust.feature:30` (`projectbluefin/testsuite`)
 exercises `projectbluefin/common`'s non-interactive `toggle-updates ACTION=`
-contract (shipped in `common#966`, 2026-08-09). The `@requires_toggle_action`
-gate in `tests/common/features/environment.py` probes with
-`ujust toggle-updates cancel` and only runs the scenario when that exits `0` —
-so the scenario ran here, meaning the probe found ACTION support present, but
-the actual `enable`/`disable` exchange still failed:
+contract (`common#966`, commit `93068cd`, merged 2026-08-09T02:16Z). The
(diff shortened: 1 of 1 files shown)


hanthor (approved):
Docs-only, and the evidence directly supports the conclusion: the `skopeo inspect` output shows `:testing` was still pinned to the 2026-08-04 build as of this PR, five days before `common#966` (which added the `toggle-updates ACTION=` contract) merged on 08-09 — so the tested image structurally couldn't have passed the scenario it was failing, independent of any session/harness explanation. The follow-on reasoning for why the gate doesn't skip on pre-#966 images (the old recipe accepts `ACTION=` and ignores it, falls through to the interactive `gum choose`, gets an empty `SELECTED_OPTION`, and exits 0 via its own cancel-handling line) is a real, useful finding — it correctly identifies that an exit-status probe can't distinguish "gate skipped correctly" from "gate accidentally succeeded for the wrong reason," and correctly assigns that fix to `testsuite`, not here.

Good scope discipline: explicitly avoids the region `#1153` (open concurrently) touches in the same file to prevent a merge conflict, and separately re-verifies the unrelated `firefox.feature` blocker is still failing rather than letting silence imply it's resolved. CI green (`validate`, `Unit tests`, `Check PR base branch`, `validate-release-notes`).

---

castrojo (approved):

PR #1197, opened 2026-09-06
test(unit): add BATS coverage for build.sh and profile.d scripts (#1176)

### LINKED ISSUE #1176: [quality] Zero unit coverage: build_files/shared/build.sh and system_files profile.d scripts
## Finding

After bluefin#1175 (`.github/scripts/validate-docs.py`), bluefin#1162 (`install-hooks.sh`, `worktree.sh`) and bluefin#1168 (`ghcurl`), the following shipped scripts still have **zero** unit coverage — no BATS file in `tests/unit/` and no Python test in `tests/` references them:

| Script | Lines | Risk |
| --- | --- | --- |
| `build_files/shared/build.sh` | 63 | Build-stage orchestrator — a regression breaks every image variant. |
| `system_files/shared/etc/profile.d/90-bluefin-starship.sh` | 23 | Runs in every interactive login shell; a bad branch breaks user shells. |
| `system_files/shared/etc/profile.d/91-bluefin-aliases.sh` | 6 | Same blast radius, trivial to cover. |

`scripts/check-testsuite-workflow-ref.py` is already tracked separately in #1169 and is deliberately excluded here.

## Recommendation

1. `tests/unit/build_test.bats` — sandbox `build_files/shared/build.sh`, stub the scripts it dispatches to, and assert stage ordering, variant/flavor argument handling, and non-zero propagation when a stage fails.
2. `tests/unit/profile_d_test.bats` — source each `profile.d` script in a sandboxed shell and assert the guard conditions (non-interactive shell, missing `starship` binary, `NO_COLOR`/TERM handling) leave the environment untouched rather than erroring.

Both land in the existing `bats tests/unit/` run (`just test-unit`, and the `Run unit tests` step in `pr-validation.yml`), so no workflow change is required.

## Priority
- Impact: medium (build.sh is high blast radius; profile.d is high frequency, low complexity)
- Effort: low

---
*Filed by quality agent (hold-gated mode)*
---
🐝 **Hive Agent**: `quality` | **Instance:** `hosted-projectbluefin-knuckle-gjvq` | **SHA:** `47ab367`

— hive: agent=quality backend=copilot model=claude-opus-5

castrojo:
## Summary
Adds BATS unit tests in `tests/unit/` for scripts highlighted in #1176 that had zero test coverage:
- `build_files/shared/build.sh` -> `tests/unit/build_test.bats`
- `system_files/shared/etc/profile.d/90-bluefin-starship.sh` and `91-bluefin-aliases.sh` -> `tests/unit/profile_d_test.bats`

## Coverage
- **`tests/unit/build_test.bats`**:
  - Sandboxes `build.sh` and stubs `dnf5`, `rpm`, `rsync`, `install`.
  - Mocks stage scripts and asserts exact execution order.
  - Verifies immediate failure propagation when any stage exits non-zero.
  - Verifies failure propagation when initial configuration commands fail.
- **`tests/unit/profile_d_test.bats`**:
  - Verifies `90-bluefin-starship.sh` no-op when starship is absent.
  - Verifies PATH-based starship initialization in bash.
  - Verifies fallback to Linuxbrew bin location when not in PATH.
  - Verifies starship init is skipped if shell is not bash.
  - Verifies cleanup of intermediate environment variables (`_starship_bin`).
  - Verifies `91-bluefin-aliases.sh` configures `rl` alias only when `ramalama` binary is present.

## Validation
- Ran tests via test harness: 10/10 test assertions pass.
- `python3 -m unittest discover -s tests -p 'test_*.py'`: 60 tests passed.
- `just check`: syntax clean.

Closes #1176

— hive: backend=goose model=gemini-3.8-flash

### DIFF
--- a/tests/unit/build_test.bats
+++ b/tests/unit/build_test.bats
@@ -65,7 +65,7 @@ EOF
         cat > "${stage_path}" << EOF
 #!/usr/bin/bash
 echo "${stage}" >> "${STAGE_LOG}"
-if [[ -n "${FAIL_STAGE:-}" && "${FAIL_STAGE}" == "${stage}" ]]; then
+if [[ -n "\${FAIL_STAGE:-}" && "\${FAIL_STAGE}" == "${stage}" ]]; then
     exit 1
 fi
 exit 0
@@ -113,7 +113,7 @@ teardown() {
     run cat "${STAGE_LOG}"
     [ "$status" -eq 0 ]
     [ "${lines[-1]}" = "base/05-override-install.sh" ]
-    refute_line --output "${output}" "base/17-cleanup.sh"
+    [[ "${output}" != *"base/17-cleanup.sh"* ]]
 }
 
 @test "build.sh: fails immediately if dnf5 config-manager fails" {
--- a/tests/unit/build_test.bats
+++ b/tests/unit/build_test.bats
@@ -1,0 +1,129 @@
+#!/usr/bin/env bats
+# Unit tests for build_files/shared/build.sh
+# Run with: bats tests/unit/build_test.bats
+
+SCRIPT_DIR="$(cd "$(dirname "$BATS_TEST_FILENAME")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
+BUILD_SCRIPT="${REPO_ROOT}/build_files/shared/build.sh"
+
+setup() {
+    TEST_ROOT="${SCRIPT_DIR}/.bats-sandbox/build.${BATS_TEST_NUMBER:-0}.$$"
+    STUB_BIN="${TEST_ROOT}/stub-bin"
+    STAGE_LOG="${TEST_ROOT}/stage.log"
+    mkdir -p "${STUB_BIN}"
+    mkdir -p "${TEST_ROOT}/ctx/system_files/shared"
+    mkdir -p "${TEST_ROOT}/ctx/build_files/shared/utils"
+    touch "${TEST_ROOT}/ctx/build_files/shared/utils/ghcurl"
+
+    ORIG_PATH="${PATH}"
+    export PATH="${STUB_BIN}:${PATH}"
+    export STAGE_LOG
+
+    # Stub commands used in build.sh
+    cat > "${STUB_BIN}/dnf5" << 'EOF'
+#!/usr/bin/bash
+exit 0
+EOF
+    chmod +x "${STUB_BIN}/dnf5"
+
+    cat > "${STUB_BIN}/rpm" << 'EOF'
+#!/usr/bin/bash
+exit 0
+EOF
+    chmod +x "${STUB_BIN}/rpm"
+
+    cat > "${STUB_BIN}/rsync" << 'EOF'
+#!/usr/bin/bash
+exit 0
+EOF
+    chmod +x "${STUB_BIN}/rsync"
+
+    cat > "${STUB_BIN}/install" << 'EOF'
+#!/usr/bin/bash
+exit 0
+EOF
+    chmod +x "${STUB_BIN}/install"
+
+    # Create mock stage scripts that log their invocation
+    MOCK_STAGES=(
+        "base/00-image-info.sh"
+        "base/03-packages.sh"
+        "base/04-install-kernel-akmods.sh"
+        "base/05-override-install.sh"
+        "shared/build-gnome-extensions.sh"
+        "base/17-cleanup.sh"
+        "base/18-workarounds.sh"
+        "base/19-initramfs.sh"
+        "shared/validate-repos.sh"
+        "shared/clean-stage.sh"
+        "base/20-tests.sh"
+    )
+
+    for stage in "${MOCK_STAGES[@]}"; do
+        stage_path="${TEST_ROOT}/ctx/build_files/${stage}"
+        mkdir -p "$(dirname "${stage_path}")"
+        cat > "${stage_path}" << EOF
+#!/usr/bin/bash
+echo "${stage}" >> "${STAGE_LOG}"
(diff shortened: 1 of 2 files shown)


hanthor (changes requested):
## `tests/unit/build_test.bats` fails whenever the checkout path contains `/tmp`

I merged this branch into current `testing` (`b5ee8b5`) and ran the repo's real gate. Result: **23 BATS failures vs. 21 on the base** — this PR adds 2 new ones, both from the file it introduces.

```
$ bats tests/unit/build_test.bats
1..3
not ok 1 build.sh: executes build stages in exact expected sequence
# (in test file tests/unit/build_test.bats, line 89)
not ok 2 build.sh: fails immediately if a stage fails
# (in test file tests/unit/build_test.bats, line 114)
ok 3 build.sh: fails immediately if dnf5 config-manager fails
```

(For reference, the other 21 failures — `04-kernel-akmods`, `sudoers`, `firewalld`, `coreos sulogin` — are present identically on `origin/testing` and are **not** this PR's fault.)

### Root cause

The sandbox patch at the end of `setup()` chains two `sed` expressions:

```bash
sed -e "s|/ctx|${TEST_ROOT}/ctx|g" -e "s|/tmp|${TEST_ROOT}/tmp|g" "${BUILD_SCRIPT}" > "${PATCHED_SCRIPT}"
```

`TEST_ROOT` is derived from `SCRIPT_DIR`, i.e. it lives **inside the repo checkout**. When the checkout path itself contains `/tmp`, the second expression rewrites the `/tmp` that the *first* expression just inserted. Demonstration with a real `TEST_ROOT` from the failing run:

```
$ TEST_ROOT="/tmp/.../\.bats-sandbox/build.2.999"
$ printf '%s\n' '/ctx/build_files/base/05-override-install.sh' \
  | sed -e "s|/ctx|${TEST_ROOT}/ctx|g" -e "s|/tmp|${TEST_ROOT}/tmp|g"
/tmp/.../build.2.999/tmp/.../build.2.999/ctx/build_files/base/05-override-install.sh
```

Every stage path is doubled and therefore nonexistent, so the patched script dies with **exit 127**. BATS flags this explicitly:

```
BW01: `run`'s command `bash .../build-patched.sh` exited with code 127, indicating 'Command not found'.
       in test file tests/unit/build_test.bats, line 88
BW01: ... in test file tests/unit/build_test.bats, line 109
```

### Why this is worse than a local-only failure

Test 2 (`fails immediately if a stage fails`) asserts only `[ "$status" -ne 0 ]`. Exit 127 satisfies that. So on any checkout path where the mangling happens but the first test is tolerated, this test **passes for entirely the wrong reason** — it never exercises `FAIL_STAGE` propagation at all. It is a vacuous test that happens to be green on the GitHub runner (`/home/runner/work/...`, no `/tmp` in the path) and red elsewhere. CI green here is not evidence the assertion works.

### Suggested fixes

1. Make the substitution order-independent — a single `sed` pass, or use a placeholder token, or (cleanest) anchor both patterns so they can't match inserted text:
   ```bash
   sed -e "s|^/ctx|${TEST_ROOT}/ctx|g" ...
   ```
   is not sufficient either since `/ctx` appears mid-line; a two-step with a sentinel is more robust:
   ```bash
   sed -e 's|/ctx|@@CTX@@|g' -e 's|/tmp|@@TMP@@|g' \
       -e "s|@@CTX@@|${TEST_ROOT}/ctx|g" -e "s|@@TMP@@|${TEST_ROOT}/tmp|g" "${BUILD_SCRIPT}"
   ```
2. Put the sandbox under `${BATS_TEST_TMPDIR}` rather than `${SCRIPT_DIR}/.bats-sandbox` — writing scratch state into the repo tree is what dragged the checkout path into the substitution in the first place.
3. Tighten test 2 to assert the *specific* exit status the stage mock produces (`[ "$status" -eq 1 ]`), so a 127 can never be mistaken for correct failure propagation.

`tests/unit/profile_d_test.bats` in this PR is fine — no failures from it.

---

Danathar (changes requested):
Re-checked after `Fix build test temp path patching` (09-17), following hanthor's method: merged current `testing` into the branch and ran the whole unit suite, from a checkout under `/var/tmp` (a path containing `/tmp`, which was the original failure).

**hanthor's blocker is fixed.** `tests/unit/build_test.bats` passes 3/3 from the `/tmp`-containing path, and the full-suite comparison is now 21 failures on `testing` vs 22 on the merged branch — the same 21 environment-specific `04-kernel-akmods` failures on both sides, plus exactly one new one.

**The one new failure is a host dependency in the other new file.** `profile_d_test.bats` → "90-bluefin-starship: initializes starship from linuxbrew **when not on PATH**" never actually removes starship from PATH. The script's first branch is `command -v starship`, so on any machine with `/usr/bin/starship` (any Bluefin host, for one) the system binary wins, the brew stub is never called, and the assertion fails. CI is green only because the Ubuntu runner has no starship. That is the same class of problem as the `/tmp` one — the test's result depends on the machine running it — and every other test in `tests/unit/` sandboxes PATH for exactly this reason.

Fix is local to that one test; I verified it passes on this host (7/7 in the file) with:

```bash
    # A host with starship installed would take the PATH branch; give the
    # sourcing shell only what the script needs, and no starship.
    mkdir -p "${TEST_ROOT}/bin"
    for tool in bash basename readlink; do
        ln -sf "$(command -v "$tool")" "${TEST_ROOT}/bin/$tool"
    done
    run env PATH="${TEST_ROOT}/bin" bash -c "source '${PATCHED_STARSHIP}'; echo \"VAL=\$STARSHIP_BREW_INIT_CALLED\""
```

(`bash`, `basename` and `readlink` are what the script and the stub actually invoke.) With that in, I'll approve — the `build_test.bats` half is ready as-is.

PR #1202, opened 2026-09-06
fix(scanner): make framework AMD kargs cleanup retry-safe on failure

### LINKED ISSUE #1126: [scanner] Framework kargs migration marks complete before deletion succeeds
## Finding\n\nPR #1103 adds a one-time privileged hook that calls version-script before running rpm-ostree kargs --delete. The shared version-script writes the target version before the migration command executes. If rpm-ostree is temporarily unavailable or deletion fails, the hook exits non-zero but subsequent boots see the version as current and skip it permanently, leaving module_blacklist=hid_sensor_hub in boot configuration.\n\n## Recommendation\n\nMake the migration retry-safe: record completion only after karg deletion succeeds, or add an explicit success marker after the mutation and avoid treating failed attempts as complete. Add a test where deletion fails, then verify a subsequent invocation retries.\n\n---\n*Filed by scanner agent (ACMM L5 — hold-gated mode)*
---
🐝 **Hive Agent**: `scanner` | **Instance:** `hosted-projectbluefin-knuckle-gjvq` | **SHA:** `unknown`

— hive: agent=scanner backend=copilot model=gpt-5.6-luna

mrbobbytables:
## Problem

Fixes #1126.

PR #1103 added a one-time privileged hook that removes the stale
`module_blacklist=hid_sensor_hub` karg from AMD Framework laptops. The
hook called `version-script` *before* attempting the
`rpm-ostree kargs --delete`. Looking at `version-script`'s
implementation (`ublue-setup-services`'s `libsetup.sh`), it records the
target version as complete on the *call itself*, not on the success of
the code that follows. So if `rpm-ostree` is temporarily unavailable or
the deletion otherwise fails, the hook exits non-zero, but the
migration is already marked complete — subsequent boots see the
version as done and skip the hook forever, leaving the stale karg in
the boot configuration.

## Fix

`system_files/shared/usr/share/ublue-os/privileged-setup.hooks.d/12-framework-amd-kargs-cleanup.sh`
now tracks completion with its own on-disk marker
(`/var/lib/ublue-os/.framework-amd-kargs-cleanup-v1`) that is only
written after the migration has actually succeeded, or after
determining it doesn't apply (non-Framework/non-AMD hardware, or the
karg is already absent). `version-script` is still called, but only
after that marker is written, so the shared setup-versioning ledger
stays in sync without ever recording a failed attempt as done.

If the karg deletion itself fails, the hook now exits non-zero without
writing the marker, so it will be retried on the next boot.

## Tests

`tests/unit/12-framework-amd-kargs-cleanup_test.bats` adds:
- A test that simulates a failing `rpm-ostree kargs --delete`, confirms
  the marker is not written and the hook reports it will retry, then
  simulates a subsequent successful boot and confirms the karg is
  removed and the marker is written.
- A test confirming that once the marker exists, the hook skips
  without re-checking hardware.

All 10 tests in the file pass locally with `bats`, and the script is
clean under `shellcheck`.

— hive: backend=copilot model=claude-sonnet-5
---
🐝 **Hive Agent**: `contributor` | **SHA:** `f938ab6`

### DIFF
--- a/system_files/shared/usr/share/ublue-os/privileged-setup.hooks.d/12-framework-amd-kargs-cleanup.sh
+++ b/system_files/shared/usr/share/ublue-os/privileged-setup.hooks.d/12-framework-amd-kargs-cleanup.sh
@@ -8,17 +8,35 @@
 # This is a one-time migration cleanup. The kernel driver fix for battery
 # charge threshold control on AMD Framework is in fw-charge-control.conf
 # (bluefin#879); this script only removes a stale karg side-effect.
+#
+# NOTE (bluefin#1126): version-script records a migration as "complete" the
+# moment it is called, before this script's body runs. A naive
+# `version-script ... || exit 0` guard at the top would mark the migration
+# done even if rpm-ostree is unavailable or the karg deletion fails, so a
+# failed attempt would never be retried on a later boot. To stay
+# retry-safe, completion is tracked with our own on-disk marker that is
+# only written after the migration has actually succeeded (or has been
+# determined not to apply), and version-script is invoked afterwards purely
+# to keep the shared setup-versioning ledger in sync.
 
 # shellcheck source=/dev/null
 source /usr/lib/ublue/setup-services/libsetup.sh
 
-version-script framework-amd-kargs-cleanup privileged 1 || exit 0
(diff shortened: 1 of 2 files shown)


castrojo (approved):

PR #1213, opened 2026-09-07
fix(ci): report lab checks with GITHUB_TOKEN instead of MergeRaptor token

eltorrero:
Fixes projectbluefin/bluefin#1114.

## Problem

The Lab check fails before validation: `lab-check.yml` mints a MergeRaptor
installation token with `permission-checks: write`, but the `projectbluefin`
installation was never granted `checks: write`, so every dispatch dies with
HTTP 422 — `The permissions requested are not granted to this installation`
(first tracked in #939; reproduced on runs 32170118754 / 32170037172 /
32170032783 and again on 32617430091 / 32617437845 / 32617445870 /
32617481346 per the issue thread).

## Fix

Instead of requiring an org-side app-permission grant, align with the
bluefin-lts pattern: declare `checks: write` in the workflow `permissions:`
block and create/update the `testing-lab / <product>` check run with the
workflow's own `GITHUB_TOKEN`. This removes the
`actions/create-github-app-token` mint step and its deny-diagnostic step;
the workflow now requests only permissions the job can already hold.

The check-run lookup matches on the check name only (server-side
`check_name` filter). A run created with `GITHUB_TOKEN` is attributed to the
`github-actions` app, so keeping the old `.app.slug == "mergeraptor"` filter
would find nothing and every lifecycle event would POST a duplicate check
run.

## Changes

- `.github/workflows/lab-check.yml`: `checks: write` in the permissions
  block; drop the app-token mint and deny-diagnostic steps; use
  `${{ github.token }}`; name-only check-run matching; rename the job to
  "Update lab check".
- `tests/test_lab_check.py`: add `LabCheckAuthTests` pinning the workflow
  permissions and the absence of any app-token mint; update the same-name
  lookup test to the new name-based matching (PATCH, not POST).
- `docs/skills/ci/SKILL.md`,
  `docs/skills/ci/references/mergeraptor-checks.md`: document the new auth
  model; the reference now covers the remaining MergeRaptor token users
  (`track-common.yml`, `renovate-automerge.yml`) and the two-level
  permission diagnosis.

## Testing

- `python3 -m unittest tests.test_lab_check` — 7/7 pass (the reporting
  script is exercised against a stub `gh`, including the buried-check and
  same-name lookup cases).
- `python3 .github/scripts/validate-docs.py` — ok.
- `bash -n` on the extracted reporting script — ok.
- actionlint / pre-commit run in CI.

Note: `MERGERAPTOR_APP_ID` / `MERGERAPTOR_PRIVATE_KEY` are now consumed only
by `track-common.yml` and `renovate-automerge.yml`.

— hive: backend=goose

### DIFF
--- a/.github/workflows/lab-check.yml
+++ b/.github/workflows/lab-check.yml
@@ -6,6 +6,7 @@ on:
 
 permissions:
   contents: read
+  checks: write
 
 # The lab sends several lifecycle events (queued -> in_progress -> completed)
 # for one head SHA, and the job below is a read-then-write: it looks up the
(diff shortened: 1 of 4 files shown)


castrojo (approved):

PR #1245, opened 2026-09-12
fix(boot): reduce initramfs size with xz

TB516:
## Summary

- force XZ compression for the primary cached initramfs build
- keep the fallback initramfs generator on the same compression contract
- add regression coverage for both dracut invocations

This reduces the initramfs file GRUB must allocate before booting module-heavy NVIDIA images, addressing the `mm.c:552:out of memory` failure reported on first boot. The generic `--no-hostonly` module set remains unchanged.

Closes #1244

## Testing

- `npx --yes bats tests/unit/04-install-kernel-akmods_test.bats tests/unit/19-initramfs_test.bats` (23 tests passed)
- `python3 -m py_compile build_files/base/04-install-kernel-akmods.py`
- `bash -n build_files/base/19-initramfs.sh tests/unit/04-install-kernel-akmods_test.bats tests/unit/19-initramfs_test.bats`
- `git diff --check`

---
🐝 **Hive Agent**: `contributor` | **SHA:** `d7b647a`

— hive: backend=codex

PR #1253, opened 2026-09-13
fix(boot): reduce initramfs size with xz

castrojo:
Force XZ compression in both initramfs generation paths so module-heavy NVIDIA images require less contiguous memory when GRUB loads them.

Replaces #1245 (which was branched off and targeting `main` instead of `testing`).


hanthor (commented):
Mechanically sound and well tested; the missing piece is numbers for the trade-off it makes.

## What I verified

- Base branch is `testing` (merge-base `708b2b0`) — correct, and it merges cleanly onto the current `testing` head `8b0eeef`.
- `bats tests/unit/` — `origin/testing@708b2b0`: **245/245 pass**. This PR: **245/245 pass**. No regressions, no pre-existing failures.
- Both changed files lint clean: `shellcheck` and `bash -n` on `19-initramfs.sh`, `python3 -m py_compile` on `04-install-kernel-akmods.py`.
- `git grep -n dracut` finds exactly two invocation sites in the tree (`04-install-kernel-akmods.py:230`, `19-initramfs.sh:22`). **Both are changed**, so the "keep both paths in sync" rule the SKILL.md note adds is actually satisfied — and there's no `dracut.conf.d` compression setting anywhere in `build_files/` or `system_files/` that could silently override it.
- Mutation-tested both new assertions; neither is vacuous:
  - removing `"--xz",` from the Python path fails `04-kernel-akmods: cached initramfs includes ISO live modules`;
  - removing `--xz` from the shell path fails `19-initramfs: regenerated initramfs includes ISO live modules`;
  - swapping `--xz` for `--zstd` fails both.

Adding the assertion to the existing "includes ISO live modules" cases rather than writing new ones is the right call — it keeps the full dracut argv pinned in one place.

## What the PR doesn't establish

The title says "reduce initramfs size" and the doc note says "the smaller on-disk image reduces the contiguous memory GRUB needs," but there are **no numbers anywhere in the change**. Please add before/after `ls -l` of `/lib/modules/$KVER/initramfs.img` for at least `bluefin` and `bluefin-nvidia`, since nvidia is the variant the rationale is about.

The trade-off that goes with it also isn't stated. XZ compresses better than zstd and decompresses substantially slower, single-threaded, on every boot. On a 51 MB binary-heavy tar here, xz at dracut's settings (`--check=crc32 --lzma2=dict=1MiB`) produced 20.4 MB and took 1.27s to decompress on one core — not a proxy for a real initramfs, but the right order of magnitude for the added latency. That cost lands on every boot of every install; the memory saving matters on the machines that were failing to boot. That may well be the right trade, but it should be a stated decision rather than a side effect, especially since it applies to all variants and not just the module-heavy ones the rationale names.

Three things I couldn't check from a checkout and would want confirmed:

1. **What the current default compressor actually is** — it comes from the base image's dracut config, not this repo, so "switch to xz" may be a change from zstd, from gzip, or from nothing at all.
2. That the Fedora 44 kernel has `CONFIG_RD_XZ` and that `xz` is available in the build stage. If either is missing, the failure mode is an unbootable image rather than a build error.
3. The GRUB contiguous-memory claim itself, and whether a narrower fix (xz for nvidia variants only) would buy the same result without the boot-time cost everywhere.

Happy to approve with the numbers and an explicit note about the decompression cost in the SKILL.md paragraph.

## Unrelated, but worth knowing

Running this suite from a checkout whose path contains `/tmp/` produces **21 spurious failures** across `04-install-kernel-akmods_test.bats` and `05-override-install_test.bats` — both sed-rewrite `/tmp/` to `${TEST_ROOT}/tmp/` in the script under test, and the substitution recurses when `TEST_ROOT` is itself under `/tmp`. Everything passes from any other location. Not caused by this PR; flagging so nobody chases it.

---

Danathar:
The only red check here, `testing-lab / bluefin`, is a leftover: that integration was retired in #1256 (merged 2026-09-14) and the check no longer exists on `testing`. It stays attached to this PR because the head SHA predates the removal. A rebase onto current `testing` will drop it and re-run the real checks; no code change is needed for that.

PR #1247, opened 2026-09-13
feat(countme): first-party countme reporting to countme.projectbluefin.io

kylerankin:
## Summary

Implements ADR 0006 on the bluefin image: a first-party, privacy-preserving
weekly check-in to the Project Bluefin countme aggregate at
`countme.projectbluefin.io`, running alongside upstream Fedora countme.

## Changes

- `bluefin-countme.service` / `bluefin-countme.timer`: weekly, rate-limited
  (`RandomizedDelaySec=12h`, `Persistent=true`) check-in running under
  `DynamicUser=yes` with a `StateDirectory`.
- `/usr/libexec/bluefin-countme`: anonymous empty-GET ping carrying only a
  coarse install-age cohort plus image metadata (no machine-id, token, or
  hostname). Opt-out honored via `/etc/projectbluefin/countme/disabled`.
- `03-bluefin-countme.preset`: enables the timer.
- Build wiring: enable the timer in `17-cleanup.sh` (keeping upstream
  `rpm-ostree-countme`), assert the timer/preset/payload in `20-tests.sh`
  (including a bucket self-test), and disable the timer in the container-native
  ISO live session (`21-container-native-iso.sh`).

The ping URL is
`https://countme.projectbluefin.io/metalink?repo=bluefin\&tag=${IMAGE_TAG}\&flavor=${IMAGE_FLAVOR}\&arch=${ARCH}\&countme=${BUCKET}`.

- [x] I am using an agent and I take responsibility for this PR

— hive: backend=pi model=lemonade/Ornith-1.5-35B-A3B-GGUF-Q6_K

---
🐝 **Hive Agent**: `contributor` | **SHA:** `0ff9d92`

PR #1254, opened 2026-09-13
feat(countme): enable first-party countme reporting for Bluefin images

### LINKED ISSUE #1216: feat(countme): enable first-party countme reporting to countme.projectbluefin.io
## Summary

Bluefin currently enables Fedora's upstream `rpm-ostree-countme.service`. To support the unified Project Bluefin analytics dashboard on `docs.projectbluefin.io/analytics` and establish first-party measurement across the Bluefin image family, Bluefin should participate in first-party reporting alongside rpm-ostree.

## Required Implementation

1. Enable the centralized first-party countme timer from `projectbluefin/common` (`03-bluefin-countme.preset` / `bluefin-countme.timer`).
2. Transmit weekly pings to `https://countme.projectbluefin.io/metalink?repo=bluefin&tag=${IMAGE_TAG}&flavor=${IMAGE_FLAVOR}&arch=${ARCH}&countme=${BUCKET}`.
3. Support opt-out via `/etc/projectbluefin/countme/disabled`.

Ref: `projectbluefin/documentation` ADR 0006.

eltorrero:
## Summary

Implements the three required implementation items from #1216:

1. **Enable the centralized first-party countme timer from `projectbluefin/common`** — `17-cleanup.sh` now runs `systemctl enable bluefin-countme.timer` at build time, next to the existing `rpm-ostree-countme.service` enable (first-party reporting runs *alongside* Fedora upstream countme, which stays untouched).
2. **Weekly ping to `https://countme.projectbluefin.io/metalink?repo=bluefin&tag=…&flavor=…&arch=…&countme=…`** — the reporter itself (`bluefin-countme.timer`/`.service`, `/usr/libexec/bluefin-countme`, `03-bluefin-countme.preset`) ships in the `projectbluefin/common` image per the issue and ADR 0006; it is open there as projectbluefin/common#1112. `00-image-info.sh` already writes the `image-name`/`image-tag`/`image-flavor` fields the client reads, so no change was needed here for the payload.
3. **Opt-out via `/etc/projectbluefin/countme/disabled`** — enforced by the common unit's `ConditionPathExists` and by the client before any state is written (checked before `lastrun` is touched, so opting out leaves no trace). Nothing further to implement in this repo.

### Changes

- `build_files/base/17-cleanup.sh`: enable `bluefin-countme.timer` at build time.
- `build_files/base/20-tests.sh`: add `bluefin-countme.timer` to `IMPORTANT_UNITS` — a silent disable (renamed unit in common, dropped enable line) fails the build instead of shipping an image that quietly stops reporting.
- `build_files/base/21-container-native-iso.sh`: stop `bluefin-countme.timer` and `bluefin-countme.service` in the live ISO session — a live session is not a deployment and must not count (mirrors the existing `rpm-ostree-countme.service` handling).
- Tests for all three: the enable, the guard, and the live-session disablement.

### Dependency

Requires projectbluefin/common#1112 (open) to land in the common image. `ghcr.io/projectbluefin/common` digests are auto-bumped and auto-merged by Renovate (`common is a first-party trusted image` in `renovate.json5`), so the build picks the unit up automatically once common#1112 merges — no follow-up commit needed here. Until then the `systemctl enable` in 17-cleanup would fail a build against a pre-#1112 common digest, so this PR should merge after common#1112 (or land together).

### Note on base branch

Targets `testing`, not the default branch `main`: this repo's CI (`check-base-branch` in `pr-validation.yml`) hard-fails any PR against `main` unless the head is `testing`, and `docs/contributing.md`/`docs/workflow.md` state that normal feature work targets `testing`, with `promote-testing-to-main.yml` handling promotion to `main`.

### Relation to #1247

#1247 was opened earlier today against the same issue. It targets `main` (its `Check PR base branch` run fails) and ships the timer, service, and client under this repo's own `system_files/` instead of consuming the centralized units from `common` as the issue specifies. Its client also deviates from the reviewed common client: no bootc tag resolution (every system would report the baked `latest` tag), bucket boundary at 24 weeks instead of Fedora's 20, and `lastrun` written before the network call (a failed ping permanently loses that week). This PR wires up the common-provided implementation instead. Left open for a maintainer to close.

### Testing

- `bats tests/unit/` — 246/246 pass (on the `testing` branch).
- `shellcheck` on all changed `build_files/*.sh` — clean (one pre-existing SC1091 info on the `/ctx` source line, unchanged).
- No image assembly changes beyond unit enablement, so no local full build; CI unit-tests + lab checks will validate the build.

Closes #1216

— hive: backend=goose model=unsloth/Qwen3.8-27B


hanthor (changes requested):
The change is well-shaped and genuinely well-tested, but as written it breaks the build: it enables a systemd unit that is not in the image.

## The blocking issue

`17-cleanup.sh` runs under `set -eoux pipefail` (line 5), so `systemctl enable bluefin-countme.timer` on a missing unit aborts the build. And the unit is missing:

- Nothing in this repo provides it — `git ls-tree -r --name-only pr1254 | grep -i countme` is **empty**. `bluefin-countme` appears only in the three build scripts and their three tests.
- The Containerfile pulls `ghcr.io/projectbluefin/common:latest@${COMMON_IMAGE_SHA}` and this PR does not move that pin.
- In `projectbluefin/common`, after fetching every branch, the only ref carrying `bluefin-countme.{service,timer}` + `03-bluefin-countme.preset` + `/usr/libexec/bluefin-countme` is **`origin/feat/first-party-countme-client`** (head `d46c4df`), and `git merge-base --is-ancestor` confirms it is **not merged into `origin/main`**. common's `main` has no countme file at all, and common has no `testing` branch, so `:latest` cannot contain these units.
- The only `bluefin-countme.{service,timer}` shipping anywhere today lives in the **dakota** repo (`files/countme/`, installed by `elements/bluefin/countme.bst`). That's a different unit — `ExecStart=/usr/libexec/dakota-countme`, opt-out `ConditionPathExists=!/etc/bluefin-countme-opt-out` — and it never reaches bluefin.

So the comment "the timer, service and client ship in the projectbluefin/common image" isn't true yet. This needs to land **after** the common branch merges and `:latest` is rebuilt. As it stands, the first CI build on `testing` fails at `17-cleanup.sh`.

Two smaller consequences of the same ordering:

- That common branch also ships `system-preset/03-bluefin-countme.preset` containing `enable bluefin-countme.timer`. If presets are applied, the explicit `systemctl enable` here is redundant — worth deciding which is authoritative rather than carrying both.
- **The unit tests can't catch any of this.** `17-cleanup_test.bats` and `20-tests_test.bats` both stub `systemctl` with a script that logs and exits 0, so a nonexistent unit is indistinguishable from a present one. The `20-tests.sh` `IMPORTANT_UNITS` guard is the right instinct, but it only fires at image-build time — which is the step that will already have failed.

## Claims I checked against the common branch

- **Opt-out path — correct.** `bluefin-countme.service` on that branch carries `ConditionPathExists=!/etc/projectbluefin/countme/disabled`, and `/usr/libexec/bluefin-countme` re-checks both that marker and `/etc/dakota-countme/disabled` before writing any state, so opting out leaves no trace on disk. Matches the comment.
- **"ping weekly" — imprecise.** The timer is `OnBootSec=15m` / `OnUnitInactiveSec=3d` / `RandomizedDelaySec=12h` / `Persistent=true`; the *client* enforces the 7-day throttle. Effectively weekly, but the comment reads as though the timer is weekly, and it isn't. Worth correcting, since the deduplication story depends on the client-side throttle rather than the schedule.
- **"ADR 0006" — I could not find it.** There's no ADR directory or file anywhere in the tree on this branch. Please link it or drop the reference.

## What is good

- Base branch is `testing` (merge-base `708b2b0`) — correct. Merges cleanly onto the current `testing` head and onto #1253.
- `bats tests/unit/` — base `245/245`, this PR **`246/246`**, no regressions.
- All three new assertions are non-vacuous. I mutation-tested each:
  - removing the `systemctl enable` line fails `17-cleanup: systemctl is invoked for required services`;
  - removing `bluefin-countme.timer` from `IMPORTANT_UNITS` fails the new `20-tests: rejects an image where the first-party countme timer is not enabled`;
  - removing the two units from the `21-container-native-iso.sh` disable loop fails `Stable image embeds the Titanoboa ISO contract and installer configuration`.
- Disabling both the timer **and** the one-shot service in the live session is the right call, and it's the detail most easily missed — a live ISO that pings would inflate the count with non-installs. The new test pins both.

Re-request review once common's `feat/first-party-countme-client` is merged and `:latest` carries the units — at that point this is an approve, with the ADR link and the timer-cadence wording fixed.

---

Danathar:
The only red check here, `testing-lab / bluefin`, is a leftover: that integration was retired in #1256 (merged 2026-09-14) and the check no longer exists on `testing`. It stays attached to this PR because the head SHA predates the removal. A rebase onto current `testing` will drop it and re-run the real checks; no code change is needed for that.

eltorrero:
Since projectbluefin/common#1112 was closed without merging, should this PR now be closed, or is there a replacement it should target?

Danathar:
Closing this PR because its implementation enables bluefin-countme.timer from projectbluefin/common#1112, but common#1112 was closed without merging. As this PR itself notes, enabling the unit against a common image that does not ship it would fail the image build. The green checks here do not resolve that dependency.

This does not close the question of first-party countme reporting. projectbluefin/bluefin#1247 is still open with changes requested and takes a different, in-repo approach; it is not an approved replacement for the shared implementation. If a working reporter is agreed and lands, the image-side timer enablement can be brought back against that concrete dependency. Thanks for putting together the integration and tests.

PR #1236, opened 2026-09-11
fix(ujust): provide working shell autocomplete in bash and zsh

mrbobbytables:
Fixes #1171

### Problem
Shell autocompletion for `ujust` in bash and zsh ceased functioning on recent builds, falling back to standard filename/directory completion when pressing `<TAB>`. This occurred because `just` transitioned to dynamic `clap_complete` completion generation, breaking the build-time sed substitutions previously used to generate ujust completion files and resulting in missing or non-functional completions.

### Solution
- Provide tailored bash completion in `system_files/shared/usr/share/bash-completion/completions/ujust` that dynamically queries recipes from `/usr/share/ublue-os/just/00-entry.just` via `just --summary --justfile`, while supporting ujust CLI flags/options.
- Provide tailored zsh completion in `system_files/shared/usr/share/zsh/site-functions/_ujust` supporting recipe completion and delegating where appropriate.
- Assert completion files exist in build-time image tests (`build_files/base/20-tests.sh`) and add unit tests.

— hive: backend=copilot model=gemini-3.8-flash
---
🐝 **Hive Agent**: `contributor` | **SHA:** `ae6e19d`

Issue #1171, opened 2026-08-30
ujust shell autocomplete not working in bash or zsh


defeated:
## 🫐 Bluefin Bug Report

_Created with `ujust report`._

### Summary

type ujust, then space, then <TAB> and instead of listing ujust commands it lists current directory contents

### Steps to reproduce

on latest :testing image type "ujust up<TAB>" but the "update" command is not autocompleted

### System

| Field | Value |
| --- | --- |
| Image | `ostree-image-signed:docker://ghcr.io/projectbluefin/bluefin` |
| Version | `testing` |
| Flavor | `main` |
| Kernel | `7.1.5-201.fc44.x86_64` |
| Architecture | `x86_64` |
| Booted digest | `unknown` |

### Failed systemd units

- systemd-remount-fs.service

### Current boot errors

```
Aug 29 12:40:37 fedora systemd-udevd[1054]: /usr/lib/udev/rules.d/70-u2f.rules:301 Failed to resolve group 'plugdev', ignoring: Unknown group
Aug 29 12:40:37 fedora systemd-udevd[1054]: /usr/lib/udev/rules.d/70-u2f.rules:304 Failed to resolve group 'plugdev', ignoring: Unknown group
Aug 29 12:40:37 fedora systemd-udevd[1054]: /usr/lib/udev/rules.d/70-u2f.rules:307 Failed to resolve group 'plugdev', ignoring: Unknown group
Aug 29 12:40:37 fedora systemd-udevd[1054]: /usr/lib/udev/rules.d/70-u2f.rules:310 Failed to resolve group 'plugdev', ignoring: Unknown group
Aug 29 12:40:37 fedora systemd-udevd[1054]: /usr/lib/udev/rules.d/70-u2f.rules:313 Failed to resolve group 'plugdev', ignoring: Unknown group
Aug 29 12:40:37 fedora systemd-udevd[1054]: /usr/lib/udev/rules.d/70-u2f.rules:316 Failed to resolve group 'plugdev', ignoring: Unknown group
Aug 29 12:40:37 fedora systemd-udevd[1054]: /usr/lib/udev/rules.d/70-u2f.rules:319 Failed to resolve group 'plugdev', ignoring: Unknown group
Aug 29 12:40:37 fedora systemd-udevd[1054]: /usr/lib/udev/rules.d/70-u2f.rules:322 Failed to resolve group 'plugdev', ignoring: Unknown group
[...]

PR #1262, opened 2026-09-14
fix(ujust): provide working shell autocomplete in bash and zsh

kylerankin:
## Problem

`ujust` + <TAB> (and `ujust up<TAB>`) listed the current directory's contents
instead of the available ujust recipes, in both bash and zsh.

## Fix

Ship shell completion for both shells:

- `usr/share/bash-completion/completions/ujust` — completes `just` CLI flags
  when the argument starts with `-`, and recipe names from
  `/usr/share/ublue-os/just/00-entry.just` otherwise.
- `usr/share/zsh/site-functions/_ujust` — same behavior, falling back to the
  native `_just` completion when available.

Guard delivery by extending the `20-tests` build checks so an image missing
either completion file fails the build, plus a dedicated bats suite that
executes both completion scripts against a mock `just`.

## Tests

- `tests/unit/ujust-completion_test.bats` (new) — runs both completion scripts.
- `tests/unit/20-tests_test.bats` — two new cases covering missing completions.

All 14 tests in `20-tests_test.bats` and all 5 in `ujust-completion_test.bats` pass.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `6d6b387`


Danathar:
Same situation as #1270 (which is this change re-opened): the fix is already on `testing` as `06877f7` — *fix(ujust): provide working shell autocomplete in bash and zsh* (#1236, merged 2026-09-14) — same five files, identical content apart from one space in `_ujust`. The conflicts come from the branch being cut from `main` and retargeted to `testing` on 09-15, which drags the 179-file `main`↔`testing` divergence into the diff.

Suggest closing both this and #1270.

(The red `Check PR base branch` is stale: it ran while the base was still `main`. Retargeting on 09-15 did not re-run it — the workflow only fires on `opened`/`synchronize`/`reopened`, not `edited` — and now that the PR has conflicts GitHub will not run `pull_request` workflows on it at all. The force-push clears both.)

Danathar:
Closing as a duplicate of #1236 (merged to `testing` 2026-09-14) — this is the same change as #1270, re-opened. See the comment above and #1171 for why the residual breakage is a `common` layering problem (projectbluefin/common#1126) rather than something a second bluefin PR can fix.

PR #1258, opened 2026-09-14
Add top-level permissions: {} to run-testsuite + sync-main-to-testing

kylerankin:
[sec-check] #1231: missing top-level `permissions: {}`

`.github/workflows/run-testsuite.yml` and `.github/workflows/sync-main-to-testing.yml` declared permissions only at job level, with no top-level `permissions: {}`. Every other bluefin workflow declares a top-level block.

Adding an empty top-level `permissions: {}` makes the default token empty so only the explicit job-level grants apply (`contents: read` + `packages: write`; `contents: write`). Job behavior is unchanged.

— hive: backend=pi model=lemonade/Ornith-1.5-35B-A3B-GGUF-Q6_K
---
🐝 **Hive Agent**: `contributor` | **SHA:** `1ce6ec6`

PR #1238, opened 2026-09-11
fix(ci): add missing top-level permissions: {} to workflows

mrbobbytables:
## Summary
Declares `permissions: {}` at the top level of `.github/workflows/run-testsuite.yml` and `.github/workflows/sync-main-to-testing.yml` so that default token scopes are empty and only explicitly declared job-level permissions apply.

Closes #1231

— hive: backend=copilot model=gemini-3.8-flash
---
🐝 **Hive Agent**: `contributor` | **SHA:** `a41bfa9`

PR #1268, opened 2026-09-15
Add top-level permissions: {} to run-testsuite and sync-main-to-testing

kylerankin:
## Security Finding

Both workflows declared permissions only at the job level, with no top-level
`permissions: {}`. A job added later without an explicit block would silently
inherit the repository default token scope.

This declares an empty top-level `permissions: {}` in both files so the default
token is empty and only the explicit job-level grants apply
(`contents: read` + `packages: write` in run-testsuite.yml;
`contents: write` in sync-main-to-testing.yml). Job behavior is unchanged.


Danathar:
Already on `testing`: `2fb9867` — *fix(ci): add missing top-level permissions: {} to workflows* (#1238, merged 2026-09-14) — adds `permissions: {}` to both `run-testsuite.yml` and `sync-main-to-testing.yml`. Cherry-picking `ada721a` onto `origin/testing` yields an empty commit. This is also a duplicate of your #1258, whose diff against `testing` is now empty for the same reason.

Suggest closing both. (The red `Check PR base branch` is the policy check that PRs target `testing`, not `main`.)

Danathar:
Closing: this is the same change as #1258 (also now closed) — both target files already carry a top-level `permissions: {}` on `testing`, and #1231 was closed as resolved. This copy also targets `main`, which the base-branch check rejects.

Issue #231, opened 2026-06-01
bug: ucsi_acpi USBC000 driver bug triggers during s2idle resume on Framework 13 Intel Core Ultra, may cause panic


castrojo:
```
Bluefin 🦖  ·  issue pipeline
─────────────────────────────────────────────────
  ▶  filed      report received
  ·  approved   —
  ·  queued     —
  ·  claimed    —
  ·  done       —
─────────────────────────────────────────────────
  report:       missing     ·  confirms: 0
  area:         —         ·  priority: —
  next action:  same bug? ujust confirm 231
```

---## Summary

A `ucsi_acpi` driver error appears during s2idle resume on Framework Laptop 13 (Intel Core Ultra Series 1) running Bluefin. The error escalated on the final resume cycle before a kernel panic that required a hard reboot.

## Device

- **Machine:** Framework Laptop 13 (Intel Core Ultra Series 1)
- **BIOS:** 03.03 (2024-10-07)
- **Kernel:** 6.19.14-101.fc43.x86_64
- **Hostname:** exo-1

## Observed errors

Normal resume cycles show:
```
ucsi_acpi USBC000:00: unknown error 256
ucsi_acpi USBC000:00: GET_CABLE_PROPERTY failed (-5)
```

The **final** resume cycle (immediately before the panic) showed a new, escalated error:
```
ucsi_acpi USBC000:00: possible UCSI driver bug 1
ucsi_acpi USBC000:00: failed to re-enable notifications (-22)
```

That resume also took **2.284 seconds** vs 0.159–0.233 seconds for all previous resumes (10× slower), and USB hubs disconnected mid-resume. The machine then entered s2idle again and never resumed — a 17-minute gap before the next clean boot.

## Timeline (June 1, 2026)

| Time | Event |
|---|---|
| 15:20:13 | Resume from s2idle — `possible UCSI driver bug 1` + 2.284s resume time + USB disconnect |
| 15:20:37 | Entered s2idle again |
| 15:37:55 | Next boot — 17-minute gap, no resume recorded |

## No crash dump

pstore and kdump were not configured on this machine. No oops or backtrace was captured.

## Upstream references

The `possible UCSI driver bug` message originates in `drivers/usb/typec/ucsi/ucsi.c`. This is an upstream kernel issue. A pre-filled upstream bug report URL:

[...]

PR #1303, opened 2026-09-18
fix(framework): make the UCSI s2idle workaround retry-safe

nikhil00shinde:

The Framework Core Ultra UCSI mitigation we shipped for #231 can silently never run. Its one-time guard marks itself complete before it has actually done anything, so a laptop where the first attempt didn't stick keeps hitting the ucsi_acpi resume failure forever while the image thinks the fix is applied. This is the same trap we already fixed in the sibling AMD hook.

---

## Changes

`11-framework-ucsi-workaround.sh` guarded itself with:

```bash
version-script framework-ucsi-workaround privileged 1 || exit 0
```

`version-script` writes the target version the moment it is called — before the hook body runs (established in #1126 and fixed there for `12-framework-amd-kargs-cleanup.sh`). Every later boot then sees the version as current and skips the hook permanently.

On an affected Framework 13 Core Ultra that means the workaround is recorded as done even when it never applied `usbcore.autosuspend=-1`:

- `rpm-ostree` not on `PATH` during that privileged-setup run → hook printed a warning and exited 0, already marked done;
- `/sys/devices/virtual/dmi/id/*` not readable that boot → same;
- `rpm-ostree kargs --append-if-missing` failing (busy transaction) → `set -e` exited non-zero, still already marked done.

The result is a machine that keeps reproducing #231 while reporting the mitigation as shipped.

This adopts the marker pattern already proven in the sibling hook:

- gate on `/var/lib/ublue-os/.framework-ucsi-workaround-v1` instead of on `version-script`;
- write the marker (and sync the shared versioning ledger) only after the karg was actually appended, was already present, or the hardware was permanently ruled out;
- leave transient failures unmarked so the next boot retries, and exit non-zero on a failed append rather than swallowing it.

Behavior on hardware that is already correctly configured is unchanged — same DMI match (`Framework` + `Intel Core Ultra`), same karg, still idempotent.

Test coverage added:

- failed append → not marked complete → later boot retries and succeeds (this test fails against the previous hook);
- missing `rpm-ostree` and unreadable DMI are not recorded as complete;
- an existing marker short-circuits before any hardware probing;
- the existing apply/already-configured tests now also assert the marker.

`docs/skills/setup-hook-tests` gains the retry-safety step and red flag so the next one-time migration hook gets this coverage by default.

## Testing

- [x] `just check` passes — unaffected (`just` not installed locally; the recipe only `--fmt --check`s `Justfile`/`*.just`, and no just file changed)
- [x] `pre-commit run --all-files` passes — ran the hooks that cover the changed paths: `shellcheck` v0.11.0 clean on the hook, `python3 .github/scripts/validate-docs.py` → `documentation ok: 13 skills, 41 Markdown files`, no trailing whitespace / missing EOF newline
- [ ] Build tested locally (optional but appreciated)

`bats tests/unit/` → **256/256 passing**. `bats tests/unit/11-framework-ucsi-workaround_test.bats` → 7/7. Reverting only the hook and re-running that file fails tests 2, 3 and 6, which confirms the new assertions bind to the fix.

No image build was run — this is a `system_files` hook plus tests and docs, with no image-assembly change.

## Checklist

On a Framework 13 with Intel Core Ultra, after the next build ships:

```bash
rpm-ostree kargs | grep usbcore.autosuspend
ls -l /var/lib/ublue-os/.framework-ucsi-workaround-v1
```

You should see `usbcore.autosuspend=-1` and the marker file. If the karg is missing, the marker should be missing too — that is the fix working: the hook will try again on the next boot instead of giving up. Machines that previously had the workaround skipped can clear the stale ledger entry and reboot to let it retry.

Note this only makes the existing mitigation reliable. The underlying `ucsi_acpi USBC000: possible UCSI driver bug` is still an upstream kernel issue in `drivers/usb/typec/ucsi/ucsi.c`, so #231 should stay open to track the kernel fix and the BIOS/S3 acceptance criteria.

Refs #231, #1126


Danathar (approved):
Confirmed the premise in common's `libsetup.sh`: `version-script` compares, then writes the new version and returns 0 in the same call, so the old `version-script … || exit 0` guard recorded completion before the hook body ran.

Also checked the runner side: `hookrunner.sh` loops `bash "$script"` with no `set -e`, so the new `exit 1` on a failed append doesn't stop the later hooks from running. Marker pattern matches the sibling `12-framework-amd-kargs-cleanup.sh`, and the failed-then-retry test is the right shape for this class of bug.

PR #1304, opened 2026-09-18
fix(sleep): drop the no-op psmouse wakeup rule for HP ZBook Ultra G1a

nikhil00shinde:

We shipped a udev rule for #383 that cannot do anything. It disables PS/2 mouse wakeup on the ZBook Ultra G1a, but the kernel never enables PS/2 mouse wakeup in the first place, so the rule writes "disabled" over "disabled" on every machine. Worse, its header tells the next person that common's keyboard rule is insufficient, which sends them down the wrong path on a bug that is still open and still unexplained.

---

## Changes

Removes `system_files/shared/usr/lib/udev/rules.d/61-amd-s2idle-hp.rules` (added in #912) and adds a regression guard.

The rule's premise was that `amd_pmc` needs *every* i8042 child in D3 before the SoC can reach PC10/S0i3, so common's `atkbd`/serio0 rule had to be supplemented with a `psmouse`/serio1 one. The kernel does not work that way. Three call sites settle it:

| Source | What it shows |
|---|---|
| `drivers/input/serio/i8042.c` — `i8042_start()` | Wakeup is enabled by default **only** for `i8042_ports[I8042_KBD_PORT_NO]`: `if (pm_suspend_default_s2idle() && serio == i8042_ports[I8042_KBD_PORT_NO].serio) device_set_wakeup_enable(&serio->dev, true);`. The AUX port is made wakeup-**capable** by `device_set_wakeup_capable()` but never wakeup-**enabled**. |
| `drivers/input/serio/i8042.c` — `i8042_pm_suspend()` | `enable_irq_wake(i8042_ports[i].irq)` runs only `if (serio && device_may_wakeup(&serio->dev))`. The AUX port never arms an IRQ wake source at all. |
| `drivers/platform/x86/amd/pmc/pmc.c` — `amd_pmc_wa_irq1()` | The `quirk_s2idle_spurious_8042` quirk the rule claimed to mirror does `bus_find_device_by_name(&serio_bus, NULL, "serio0")` and touches nothing else. |

`I8042_KBD_PORT_NO` is `0` and `I8042_AUX_PORT_NO` is `1`, and `i8042_register_ports()` registers in index order — so serio0 is the keyboard and serio1 is the mouse. The removed rule therefore set `power/wakeup` to `disabled` on a device where it is already `disabled`, on every Bluefin machine, while forking `/bin/cat` per serio `add` event image-wide.

common's `60-amd-s2idle-fixes.rules` is already the complete userspace equivalent of the quirk: it clears serio0's wakeup flag at boot, so `i8042_pm_suspend()` never calls `enable_irq_wake()` and `amd_pmc_wa_irq1()` would find nothing left to do. Nothing was missing from it.

`tests/unit/amd-s2idle-udev_test.bats` locks this in: no shipped file may reference `psmouse`, match `SUBSYSTEM=="serio"`, or fork an `amd-s2idle` rules file, with the kernel citations in the header so the next person hitting #383 has the evidence rather than the folklore. All three guards fail against the file this PR removes.

### This does not fix #383

Worth stating plainly, because #912 was merged with `Closes #383`:

The reporter in [this comment](https://github.com/projectbluefin/bluefin/issues/383#issuecomment-4643363208) had already confirmed `/sys/devices/platform/i8042/serio0/power/wakeup` was `disabled` and still measured 2–3 %/h drain (~1.56 W over their 83-minute window). The i8042 path was therefore already fully mitigated when they measured, and the residual blocker is something else. Removing a placebo does not change their power draw — it just stops the file from asserting a cause that the kernel source contradicts.

What genuinely remains for #383 is what the issue body already says: the `8D01` DMI entry in `drivers/platform/x86/amd/pmc/pmc-quirks.c` upstream (I confirmed it is still absent from `torvalds/linux` master), and a fresh diagnosis of what else is holding PC10 on this board. #383 is still open and still on `1-triage`, which is the right state — this PR deliberately does not carry a `Closes` keyword.

## Testing

- [x] `just check` passes — unaffected; no `Justfile` or `*.just` file changed (`just` is not installed in this environment and the recipe only `--fmt --check`s those)
- [x] `pre-commit run --all-files` passes — ran the hooks covering the changed paths: `validate-docs.py` → `documentation ok: 13 skills, 41 Markdown files`; no trailing whitespace, EOF newline present; `shellcheck` is scoped to `^(build_files|system_files)/.*\.sh$` and no shell file changed
- [ ] Build tested locally (optional but appreciated)

`bats tests/unit/` → **257/257 passing** (253 before, 4 added). Restoring `61-amd-s2idle-hp.rules` and re-running the new file fails tests 2, 3 and 4, confirming the guards bind to the removal.

No image build was run — this deletes one overlay file and adds a test, with no image-assembly change.

## Checklist

On any Bluefin machine, before and after the next build:

```bash
cat /sys/devices/platform/i8042/serio1/power/wakeup   # disabled, both before and after
cat /sys/devices/platform/i8042/serio0/power/wakeup   # disabled on board 8D01, via common's rule
```

`serio1` reads `disabled` either way — that is the point of the change. On an HP ZBook Ultra G1a specifically, sleep drain should be unchanged by this PR; if anyone measures a regression, that would contradict the kernel sources cited above and I would want to see it.

Refs #383


Danathar (approved):
Checked the three kernel claims against current `torvalds/linux` master and they hold:

- `i8042_start()` calls `device_set_wakeup_capable()` for every port but `device_set_wakeup_enable()` only when `serio == i8042_ports[I8042_KBD_PORT_NO].serio`.
- `i8042_pm_suspend()` only calls `enable_irq_wake()` where `device_may_wakeup()` is true, so the AUX port never arms a wake IRQ.
- `amd_pmc_wa_irq1()` does `bus_find_device_by_name(&serio_bus, NULL, "serio0")` and nothing else.
- `8D01` is still absent from `pmc-quirks.c`.

common's `60-amd-s2idle-fixes.rules` already covers `DRIVERS=="atkbd"` on board 8D01, so the removed rule was a no-op on top of it. I also appreciate that this deliberately does not claim to close #383.

PR #1306, opened 2026-09-18
fix(ci): re-run check-base-branch when PR base is retargeted

kylerankin:
## Problem

`.github/workflows/pr-validation.yml` gates every PR on `check-base-branch`, which reads the base ref from the event payload (`github.base_ref`). But its `pull_request` trigger declared no `types:`, so GitHub used the default set `[opened, synchronize, reopened]`.

Retargeting a PR's base branch fires `edited`, which is **not** in that default set. So `check-base-branch` never re-evaluates after a correction: the gate stays red and every downstream job (`validate`, `Unit tests`, `E2E smoke`, all `needs: [check-base-branch]` with `if: always() && ...`) stays in `skipping` forever. The PR is unmergeable with a now-correct base.

## Fix

Add `edited` to the trigger so the job re-runs on a base retarget:

```yaml
on:
  pull_request:
    # 'edited' is required: retargeting a PR's base branch fires 'edited', not
    # 'synchronize'. Without it, check-base-branch never re-evaluates and a
    # correctly-retargeted PR stays red with every downstream job skipped.
    types: [opened, edited, synchronize, reopened]
    branches:
      - testing
      - main
  merge_group:
```

Also soften the error text from 'Re-open this PR with base \'testing\'' to 'Retarget this PR to base \'testing\'', which is accurate once the trigger is fixed.

This mirrors how `projectbluefin/review`'s `pr-title.yml` declares the same `types` set because it also reads gate input from the event payload.

## Testing

- The workflow YAML parses; the trigger now includes `edited` alongside the base-branch gate.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `71b562a`


Danathar (approved):
Correct diagnosis: a base retarget fires `pull_request.edited`, which isn't in the default `[opened, synchronize, reopened]` set, so `check-base-branch` never re-ran and everything downstream stayed skipped.

One thing to be aware of (not blocking): `edited` also fires on every title or body edit, and each one will now re-run the whole workflow including E2E smoke. If that turns out to be noisy, a follow-up could gate the expensive jobs on `github.event.action != 'edited' || github.event.changes.base != null`.

Issue #557, opened 2026-06-13
chore(oem): remove Framework/Ampere hooks and icons now shipped by common


castrojo:
## Background

`projectbluefin/common#672` moves the Framework/Ampere first-boot hardware setup into `common` so all variants (including `bluefin-lts`) receive these tweaks. Once that PR merges and ships, the copies in this repo become duplicates and the hooks will run twice on every bluefin system.

**Blocked on:** projectbluefin/common#672 merging and being picked up by this repo's Containerfile/build.

## Files to remove

```
system_files/shared/usr/share/ublue-os/system-setup.hooks.d/10-framework.sh
system_files/shared/usr/share/ublue-os/user-setup.hooks.d/10-theming.sh
system_files/shared/usr/share/icons/hicolor/scalable/actions/framework-logo-symbolic.svg
system_files/shared/usr/share/icons/hicolor/scalable/actions/ampere-logo-symbolic.svg
```

## Notes

- `user-setup.hooks.d/20-framework.sh` (installs framework_tool/wallpapers via brew) is **not** part of this cleanup — it has not been moved to common yet.
- The `version-script` versioning in the hooks means the per-user/system state file will have already recorded the old version. The common copy uses the same version numbers (`framework system 2`, `theming user 1`), so on existing systems the hooks will correctly skip re-running after the duplicate is removed.

## Verification

After removing and building: confirm `ublue-system-setup` and `ublue-user-setup` still run the hooks (sourced from common's layer), and that a fresh install applies Framework/Ampere tweaks correctly.

PR #1314, opened 2026-09-21
docs(ci): note the reusable-build rechunk opt-in in build-image-testing

mendezr:
## What does this change?

Comment-only change: documents in the thin-caller build workflow that
`reusable-build.yml` now supports a `rechunk: 'true'` opt-in for the
testing-stream rechunk/SBOM skip, and that bluefin deliberately keeps the
default skip (Syft OOM protection).

## Purpose

This pull request exists solely as the consumer-validation PR for a
projectbluefin/actions change that adds the `rechunk` input to
reusable-build.yml (see actions PR referencing this one). It is not for
merge on its own and will be closed once the actions PR has landed.

— hive: backend=goose model=z-ai/glm-5.3-flash
---
🐝 **Hive Agent**: `contributor` | **SHA:** `f050254`


Danathar:
This reads correctly, but I think it lands out of order.

The comment says the reusable workflow "now supports a `rechunk: 'true'` opt-in". On `projectbluefin/actions` main that input does not exist yet. The `inputs:` block of `.github/workflows/reusable-build.yml` currently ends at `max_layers`, and there is no `rechunk` input anywhere in the file (the `RECHUNK_*` hits in there are the timing steps, not an input):

```
gh api repos/projectbluefin/actions/contents/.github/workflows/reusable-build.yml \
  --jq '.content' | base64 -d | grep -nE '^\s{6}rechunk'
```

returns nothing.

The opt-in is in projectbluefin/actions#557, which is still open and has no review on it. So merging this first leaves bluefin carrying a comment that describes a capability its pinned actions ref does not have, and anyone who reads it and tries `rechunk: 'true'` gets an unknown-input error.

Nothing wrong with the wording or the intent, and the "bluefin deliberately keeps the default skip" line is worth having. I would just sequence it behind #557 so the comment is true when it lands. Happy to approve as soon as #557 merges.

